PR #7, opened 2014-12-03
psutil is included with the 'omnibus' install of the agent

bakins:


gvilarino:
Haven't, sorry. Will update it for sure, tho.

PR #13, opened 2016-01-21
Added python-psutil which is not included by default but needed for s…

ssbarnea:
Added psutil installation as is needed for https://github.com/DataDog/dd-agent/issues/853


olivielpeau:
Thanks @ssbarnea for opening the issue!

`psutil` is bundled with the deb and rpm `datadog-agent` package that the role uses, so we shouldn't need to install any additional package to use `psutil` in the agent.

You can check this by running the embedded python:

``` bash
$ /opt/datadog-agent/embedded/bin/python -c "import psutil; print psutil.__version__"
2.2.1
```

PR #20, opened 2016-03-22
Allow apt repo settings to be user defined

ghost:
To allow the use of local apt mirrors

### DIFF
--- a/tasks/pkg-debian.yml
+++ b/tasks/pkg-debian.yml
@@ -2,7 +2,11 @@
 - apt: name=apt-transport-https state=latest
 
 - apt_key: id=C7A7DA52 keyserver=hkp://keyserver.ubuntu.com:80 state=present
+  when: datadog_apt_key_url is not defined
 
-- apt_repository: repo='deb http://apt.datadoghq.com/ stable main' state=present update_cache=yes
+- apt_key: id=C7A7DA52 url={{ datadog_apt_key_url }} state=present
+  when: datadog_apt_key_url is defined
+
+- apt_repository: repo='{{ datadog_apt_repo }}' state=present update_cache=yes
 
 - apt: name=datadog-agent state=latest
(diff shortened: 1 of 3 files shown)


tasks/pkg-debian.yml
```diff
@@ -2,7 +2,11 @@
 - apt: name=apt-transport-https state=latest
 
 - apt_key: id=C7A7DA52 keyserver=hkp://keyserver.ubuntu.com:80 state=present
+  when: apt_key_url is not defined
```
olivielpeau: This should be `datadog_apt_key_url`

tasks/pkg-debian.yml
```diff
@@ -2,7 +2,11 @@
...
 
-- apt_repository: repo='deb http://apt.datadoghq.com/ stable main' state=present update_cache=yes
+- apt_key: url={{ datadog_apt_key_url }} state=present
```
olivielpeau: I don't think the `id` option should be removed here: the id of the key should be identical even when you pull the key from a mirror. Do you have a use case where that wouldn't be true?

tasks/pkg-debian.yml
```diff
@@ -2,7 +2,11 @@
...
+  when: datadog_apt_key_url is defined
+
+- apt_repository: repo={{ datadog_apt_repo }} state=present update_cache=yes
```
olivielpeau: There should be quotes around the value of `repo`, i.e. `repo='{{ datadog_apt_repo }}'`

olivielpeau:
Hi @geoffwright, thanks for your contribution!

I've added a few comments on your PR, let me know what you think.

Also, can you remove your last commit from your PR (9a2a29a)? The change it provides has already been merged through a separate PR.

Thanks!

tasks/pkg-debian.yml
ghost: Yes, the id will always be the same. It isn't required, but it will save downloading the key if the id is already present in the target keyring, so I'll add it back in.

olivielpeau:
Merging, thanks @geoffwright!

PR #21, opened 2016-04-12
Avoids bare variables which is deprecated

chriscannon:
In the newer versions of Ansible a deprecation warning is thrown on this line because it is using a bare variable. I have updated the line to properly call the variable.


olivielpeau:
Looks good, thanks @chriscannon!

PR #24, opened 2016-05-20
Added datadog tag to all tasks.

BenOvermyer:
Adds the `datadog` tag to all tasks. This is a relatively tiny optimization to avoid having to tag the role in a playbook.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -12,20 +12,25 @@
     owner={{ datadog_user }}
     group={{ datadog_group }}
   notify: restart datadog-agent
+  tags: [ datadog ]
 
 # DEPRECATED: Remove specific handling of the process check for next major release
 - template: src=process.yaml.j2 dest=/etc/dd-agent/conf.d/process.yaml
   when: datadog_process_checks is defined
   notify: restart datadog
+  tags: [ datadog ]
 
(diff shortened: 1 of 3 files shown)


olivielpeau:
Hi @BenOvermyer and thanks for opening this PR!

I'm not sure about adding a `datadog` tag to all the tasks of the role: IMO the tags of the whole role should be user-defined and flexible. Having to add the tag on the role in a playbook seems reasonable to me.

Let me know if that makes sense to you. I'm going to close this PR but feel free to respond if you disagree.

BenOvermyer:
I disagree because this adds a default tag; if people don't care for it, they can override it in their playbooks. Having no default means people are required to override it in order to use tag-based filtering with the Datadog role.

However, since the latter is a minimal amount of work and only a minor inconvenience, I'm willing to let the matter go.

olivielpeau:
Thanks for your feedback @BenOvermyer.

I get your point, and I agree that having a default tag on the role can reduce some duplication when the role is used in multiple playbooks. That said I don't think Ansible allows us to implement it in a maintainable way (IMO adding the tag on every task/include statement clutters the tasks, and we could easily forget to use the tag on future additional tasks of the role), so until Ansible gives us a sensible way to tag the role with default tags I'd hold off from adding this.

See this related issue on the ansible repo: https://github.com/ansible/ansible/issues/5512

PR #30, opened 2016-09-27
Trust new APT and RPM keys

olivielpeau:
Make the pkg managers trust the new keys in addition to the current keys. Similar approach as what's described in https://github.com/DataDog/chef-datadog/pull/365.
### APT
-  Straightforward with the `apt_key` module
- Allow customizing the URL with `datadog_apt_key_url_new` (for users
  with local apt mirrors)
### RPM
- Download the new key through plain HTTP for best compatibility (since
  SSL only works when the managed nodes have a recent version of python)
- Ensure the integrity of the downloaded file using a sha-256 sum. We use
  the `sha256sum` option instead of `checksum` since `checksum` was only
  added in Ansible 2.0
- The `rpm_key` module has a good idempotent behavior so no need to have
  extra steps to check whether the key is already imported.
### Testing

Tested manually w/ Ansible `1.9.2` and `2.1.1`, against Ubuntu 14.04 and CentOS 6

### DIFF
--- a/tasks/pkg-redhat.yml
+++ b/tasks/pkg-redhat.yml
@@ -1,4 +1,13 @@
 ---
+- name: Download new RPM key
+  get_url:
+    url: "http://yum.datadoghq.com/DATADOG_RPM_KEY_E09422B3.public"
+    dest: /tmp/DATADOG_RPM_KEY_E09422B3.public
+    sha256sum: 694a2ffecff85326cc08e5f1a619937999a5913171e42f166e13ec802c812085
+
+- name: Import new RPM key
+  rpm_key: key=/tmp/DATADOG_RPM_KEY_E09422B3.public state=present
+
 - name: Copy repo file into place
   template: src=datadog.repo.j2 dest=/etc/yum.repos.d/datadog.repo owner=root group=root mode=0644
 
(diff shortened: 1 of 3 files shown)


tasks/pkg-redhat.yml
```diff
@@ -1,4 +1,13 @@
 ---
+- name: Download new RPM key
+  get_url:
+    url: "http://yum.datadoghq.com/DATADOG_RPM_KEY.public.E09422B3"
+    dest: /tmp/DATADOG_RPM_KEY.e09422b3.public
```
degemer: Does this mean this file will always be there, even when the `rpm` key is already imported ?

degemer (approved):
LGTM (but I'm far from being an ansible expert).

tasks/pkg-redhat.yml
olivielpeau: Correct. We could avoid this by first checking if the key is already imported and only download/import the new key if it's not. I didn't do it because I didn't think it was worth the added complexity, but let me know what you think.

degemer: No, it's probably not worth it given the size of the key.

olivielpeau:
Just updated the URL of the new key to `DATADOG_RPM_KEY_E09422B3.public` (which will be the actual permanent URL of the key)

@degemer: could you do a quick sanity check of my change? 🙇

degemer (approved):
Awful that Github is not able to show a proper diff. Still LGTM.

jeffwidman:
Is there a reason this wasn't merged?

PR #31, opened 2016-10-24
Remove use_mount

sarimali:
Using `use_mount` in datadog.conf has been deprecated in favor of `use_mount` in disk.yaml

### DIFF
--- a/templates/datadog.conf.j2
+++ b/templates/datadog.conf.j2
@@ -8,9 +8,6 @@
 {% if datadog_config["api_key"] is not defined -%}
   api_key: {{ datadog_api_key | default('youshouldsetthis') }}
 {% endif %}
-{% if datadog_config["use_mount"] is not defined -%}
-  use_mount: {{ datadog_use_mount | default('no') }}
-{% endif %}
 
 {# These variables are free-style, passed through a hash -#}
 {% if datadog_config -%}


olivielpeau:
Thanks @sarimali for your contribution! Since `use_mount` is still supported by the Agent (even though it is deprecated), I think we shouldn't stop supporting it in this role (dropping support would technically be a breaking change).

Could you make the template take `datadog_use_mount` into account **only if** it's defined instead? What do you think?

sarimali:
@olivielpeau Good call. I pushed the requested change.

irabinovitch:
@olivielpeau anything pending on this PR?

olivielpeau:
Actually in https://github.com/DataDog/ansible-datadog/pull/31#issuecomment-268128043 I misread what this logic currently does...

What it currently does is that it checks if the `datadog_config["use_mount"]` variable is defined to decide whether to apply the `datadog_config_use_mount` variable or not (if it's not defined we try to use `datadog_config_use_mount`, and if it is defined we don't try to use `datadog_config_use_mount` given that `datadog_config["use_mount"]` is used in the config file later in the template logic).

So this code was added to support the old `datadog_config_use_mount` ansible variable and the newer `datadog_config["use_mount"]` variable at the same time.

So we should only keep commit 5a295d2 in this PR and merge it for the next major version of the role (since it's a breaking change).

@sarimali: sorry for only getting back to you now, but could you remove commit `c3de719` from your PR to only keep the first commit (5a295d2)? We'll be able to merge this for the next major.

sarimali:
I just deleted my last commit.

olivielpeau (approved):
Thanks @sarimali, looks good! Adding this to the "next major" milestone

alopezz:
Closing this based on the age of the changes.

PR #32, opened 2016-11-01
Convert to proper YAML formatting

jeffwidman:
Using separate lines of `key=value` format isn't proper YAML syntax. Yes, technically it's supported by Ansible. However, IIRC, when it's not proper YAML, it bypasses some Ansible sanity checks because those kv pairs are passed straight through to python as strings. Plus treating it as proper YAML doesn't break syntax highlighters.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -7,10 +7,10 @@
 
 - name: Create main Datadog agent configuration file
   template:
-    src=datadog.conf.j2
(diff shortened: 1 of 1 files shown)


jeffwidman:
bump

olivielpeau:
Looks good, thanks @jeffwidman! (and sorry about the delay here)

Merging

PR #34, opened 2016-11-28
Force apt cache update unless updated within the hour

ebonharme:
Unless a recent apt update has been performed, there's a real chance that installing the latest version of this package will fail. To avoid making an update when one has been performed recently, I check the cache is fairly recent.


ebonharme:
Fixed by #36

PR #44, opened 2017-02-08
adding support to install dd-agent on suse

estib:
@olivielpeau Did I miss any content from what we put together earlier?

I set `gpgcheck=1` for now, although old zypper users may want that to be configurable. Not sure how many customers will not have the `repo_gpgcheck` variable available... Thoughts?


dabcoder:
Sorry @estib we never got around to merging this, support for Suse was added later on, closing this PR then.

PR #46, opened 2017-02-22
Ssl expiry dd

joanocallaghan:
@tmcevoy14

Adding ssl expiry check


joanocallaghan:
pushed to wrong fork..

Issue #47, opened 2017-02-27
Issues with CentOS 5


holmser:
We have a few legacy servers that are still running on CentOS 5.  This brought to light a couple of issues:

1.  It appears 5 does not support sha256
```
TASK [Datadog.datadog : Download new RPM key] **********************************
fatal: [west01]: FAILED! => {"changed": false, "failed": true, "msg": "Could not hash file '/tmp/DATADOG_RPM_KEY_E09422B3.public' with algorithm 'sha256'. Available algorithms: sha1, md5"}
	to retry, use: --limit @/home/holmser/code/systems/ansible-files/playbooks/ccs-servers.retry
```
2.  CentOS 5 does not support TLS.  SSLv3 is the latest.
```
httplib.py\", line 685, in _send_output\n    self.send(msg)\n  File \"/usr/lib64/python2.4/httplib.py\", line 652, in send\n    self.connect()\n  File \"/usr/lib64/python2.4/site-packages/M2Crypto/httpslib.py\", line 47, in connect\n    self.sock.connect((self.host, self.port))\n  File \"/usr/lib64/python2.4/site-packages/M2Crypto/SSL/Connection.py\", line 156, in connect\n    ret = self.connect_ssl()\n  File \"/usr/lib64/python2.4/site-packages/M2Crypto/SSL/Connection.py\", line 149, in connect_ssl\n    return m2.ssl_connect(self.ssl, self._timeout)\nM2Crypto.SSL.SSLError: sslv3 alert handshake failure\n", "rc": 1, "results": []}
```

hush-hush:
Hi @holmser,

Centos 5 entered EOL on March 31, 2017, so this role doesn't support it anymore. We're going to update the supported platforms with #132.

PR #48, opened 2017-02-27
Fix CentOS 5 related issues

holmser:
This PR fixes the issues outlined in [Issue 47](https://github.com/DataDog/ansible-datadog/issues/47).  There is a breaking change as `sha256sum` was deprecated for `checksum` in 2.0.

* [IMPROVEMENT] sha256sum is deprecated per [Ansible documentation](http://docs.ansible.com/ansible/get_url_module.html).  SHA1 is the only universal option across all versions, so it was chosen.
* [FEATURE] Allow APT repo setting to be user defined.


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 CHANGELOG
 =========
+# 1.3.0 / 2017-2-27
```
olivielpeau: Could you remove the release date? (since it won't be the actual release date ;) )

`Unreleased` would work for instance

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
...
+
+* [IMPROVEMENT] sha256sum is deprecated per [Ansible documentation](http://docs.ansible.com/ansible/get_url_module.html).  SHA1 is the only universal option across all versions, so it was chosen.
+* [FEATURE] Allow APT repo setting to be user defined.
```
olivielpeau: I think you meant `yum repo` instead

defaults/main.yml
```diff
@@ -13,3 +13,5 @@ datadog_group: root
 
 # default apt repo
 datadog_apt_repo: "deb http://apt.datadoghq.com/ stable main"
+
+datadog_yum_repo: "https://yum.datadoghq.com/rpm/"
```
olivielpeau, line 17: you should use the value of the variable in the template: https://github.com/DataDog/ansible-datadog/blob/1.2.0/templates/datadog.repo.j2#L3-L6

meta/main.yml
```diff
@@ -3,7 +3,7 @@ galaxy_info:
   author: 'Brian Akins, Dustin Brown & Datadog'
   description: Install Datadog agent and configure checks
   license: Apache2
-  min_ansible_version: 1.6
+  min_ansible_version: 2.0
```
olivielpeau, line 6: Is there a way we could avoid increasing the minimum ansible version? Increasing it would require a major version bump on this role IMO.

tasks/pkg-redhat.yml
```diff
@@ -3,7 +3,7 @@
   get_url:
     url: "http://yum.datadoghq.com/DATADOG_RPM_KEY_E09422B3.public"
     dest: /tmp/DATADOG_RPM_KEY_E09422B3.public
-    sha256sum: 694a2ffecff85326cc08e5f1a619937999a5913171e42f166e13ec802c812085
+    checksum: sha1:19ae0b5ef027560202632581ba213702cf78be0f
```
olivielpeau: could we make the algorithm depend on the platform so that we still use a `sha256` hash for platforms that support it?

olivielpeau (changes requested):
Thanks for your contribution @Holmser!

I've added a few comments, let me know what you think

CHANGELOG.md
holmser: :+1:

CHANGELOG.md
holmser: :+1:

defaults/main.yml
holmser: fixed

irabinovitch:
@Holmser any thoughts on @olivielpeau 's feedback?

jeffwidman (changes requested):
Why is a `.pyc` file included in this PR?

Also, IMHO adding `Molecule`, while it may (or may not) be a good idea, is a significant change that should probably be in a separate PR so it can be discussed separately from the CentOS 5 issues.

dabcoder:
Closing this as per @hush-hush's comment here: https://github.com/DataDog/ansible-datadog/issues/47#issuecomment-388933544, CentOS 5 being EOL and whilst we're considering the use of Molecule for this role, it would be good to evaluate it in a separate PR as pointed out in the previous comment.

PR #50, opened 2017-03-11
Ansible lint

eplanet:
This is a proposition to activate a kind of Ansible check-style enforcement using `ansible-lint`. It currently only ignores `[EANSIBLE0010] Package installs should not use latest` since I'm not sure how you'd like to fix it.

Please share your thoughts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+---
+language: python
+python: "2.7"
+
+install:
+  - pip install ansible-lint
+
+script:
+  - ansible-lint $(find tests -name *yml) -x ANSIBLE0010
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -14,17 +14,20 @@
   notify: restart datadog-agent
 
 # DEPRECATED: Remove specific handling of the process check for next major release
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -0,0 +1,13 @@
+---
+language: python
+python: "2.7"
+
+before_install:
+  - sudo apt-get update -qq
```
olivielpeau: why this step? If it's not useful let's remove it

olivielpeau (commented):
Thanks @eplanet! This is a very welcome change.
Disabling rule `EANSIBLE0010` is a good call, we can re-enable it once #57 is addressed.

I've enabled Travis on this repo, so your next push should trigger a travis build.

.travis.yml
eplanet: Indeed! I added it because at first I thought `ansible-lint` would need `ansible` installed, but since it's not the case, let's delete it.

eplanet:
Thanks for the comments and for the good welcome of the PR :-P. I cleaned the `.travis.yml` as discussed.

olivielpeau (approved):
👌  Thanks @eplanet!

PR #54, opened 2017-03-28
Check agents

tmcevoy14:
Adding missing functionality to add Datadog check agents to /etc/ddagent/check.d similar to the existing task that adds configs to /etc/ddagent/conf.d. It's important that all configurations, including unique python checks, can be rolled out through a singular tool. 

We use this in our forked repo by adding the needed check scripts to our repo (e.g. check_something1.py) and then defining the datadog_check_agents list variable in a given playbook's vars. 

E.g. datadog_check_agents:
         - check_something1
         - check_something2


olivielpeau:
@tmcevoy14 Any reason for closing this PR? Having the ability to add custom checks with this role would be a very welcome feature

PR #60, opened 2017-04-03
[ansible-lint] Use repository name as role name for linting

olivielpeau:
The name of the current local directory is what should be used so that
ansible-lint lints the correct task files

(Using `Datadog.datadog` would only look for installed roles, not
the role in the current directory)

Follow-up to #50, cc @eplanet


eplanet:
Very good catch indeed! Shame on me... :-)

PR #68, opened 2017-05-20
Fix incorrect handler name

jeffwidman:
The handler name is `restart datadog-agent`. The process check update `notify` call was using an incorrect name, so would never actually trigger the handler.


olivielpeau (approved):
Nice catch 👍

PR #69, opened 2017-05-20
DRY up the apt-keys task using a default variable

jeffwidman:
Uses a default var instead of repeating tasks.

Note that this changes the behavior slightly such that the same keyserver will be used for fetching both the old and new keys. Open to feedback on this, just the more I thought about it, I realized that anyone paranoid enough to run their own keyserver will likely be doing it for both keys (or simply comment out the key they don't want).

### DIFF
--- a/tasks/pkg-debian.yml
+++ b/tasks/pkg-debian.yml
@@ -2,21 +2,16 @@
 - name: Install apt-transport-https
   apt: name=apt-transport-https state=present
 
-- name: Install ubuntu apt-key server
-  apt_key: id=A2923DFF56EDA6E76E55E492D3A80E30382E94DE keyserver=hkp://keyserver.ubuntu.com:80 state=present
-  when: datadog_apt_key_url_new is not defined
-
-- name: Install Datadog apt-key
-  apt_key: id=A2923DFF56EDA6E76E55E492D3A80E30382E94DE url={{ datadog_apt_key_url_new }} state=present
-  when: datadog_apt_key_url_new is defined
-
-- name: Ensure ubuntu apt-key server is present
-  apt_key: id=C7A7DA52 keyserver=hkp://keyserver.ubuntu.com:80 state=present
-  when: datadog_apt_key_url is not defined
-
-- name: Ensure Datadog apt-key is present
-  apt_key: id=C7A7DA52 url={{ datadog_apt_key_url }} state=present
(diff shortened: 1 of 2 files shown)


tasks/pkg-debian.yml
```diff
@@ -2,21 +2,16 @@
...
-  apt_key: id=C7A7DA52 keyserver=hkp://keyserver.ubuntu.com:80 state=present
-  when: datadog_apt_key_url is not defined
-
-- name: Ensure Datadog apt-key is present
-  apt_key: id=C7A7DA52 url={{ datadog_apt_key_url }} state=present
-  when: datadog_apt_key_url is defined
+- name: Ensure Datadog apt-keys are installed
+  apt_key:
+    id: "{{ item }}"
+    keyserver: "{{ datadog_apt_key_url }}"
```
olivielpeau, line 8: actually, `datadog_apt_key_url` is not a link to a `keyserver` but to an actual direct URL from which the key can be downloaded, so they're necessarily different for the 2 keys.

So instead of `with_items` you could use `with_dict` to be able to specify both keys' fingerprint and url

olivielpeau (changes requested):
Thanks @jeffwidman! I left a comment, let me know what you think

tasks/pkg-debian.yml
jeffwidman: Good point, I overlooked that one set of tasks was using the `url` param and the other the`keyserver` param.

It's late, so maybe I'm missing something, but even when using `with_dict`, I'm not seeing how this could be DRY'd up and still retain the full logic of using a `keyserver` only for the keys where the custom `url` isn't specified?

How do you handle the case where someone specifies a custom url for one of the keys but not the other? Or can we assume it's all-or-nothing that for both keys they'll either use the keyserver or specify custom urls?

olivielpeau: You're right, `with_dict` wouldn't work since we're using a different parameter depending on whether a custom url is provided.

We could assume that if a custom url is provided for one key then the second is too, but it'd probably make the logic more complex and difficult to understand...

So I actually don't see how to DRY this up in a simple way. Maybe we could simply keep the code as it is and improve the comments/names of the tasks?

jeffwidman: I think best to just leave as-is for now, this can be simplified once you rotate the keys and don't need two copies of them.

PR #70, opened 2017-05-23
update_cache=yes in pkg-debian to make sure it doesn't bork

jbergquist:
Updated the apt-transport-https to make sure it doesn't throw an error if your cache is to damn old... like mine was.

### DIFF
--- a/tasks/pkg-debian.yml
+++ b/tasks/pkg-debian.yml
@@ -1,6 +1,6 @@
 ---
 - name: Install apt-transport-https
-  apt: name=apt-transport-https state=present
+  apt: name=apt-transport-https state=present update_cache=yes
 
 - name: Install ubuntu apt-key server
   apt_key: id=A2923DFF56EDA6E76E55E492D3A80E30382E94DE keyserver=hkp://keyserver.ubuntu.com:80 state=present


jeffwidman:
How was it breaking? Was it your cache or the package itself? 

Updating the cache without changing the value of `state=present` won't upgrade an already installed, old version of the package...

jbergquist:
Well, I ran the playbook on a machine that hadn't done an apt update for a while. So the package had been updated, the version I had in the cache had been removed so I got an 404.

Adding update_cache=yes solved my issue.

jeffwidman (approved):

olivielpeau:
Hi @jbergquist, thanks for the PR and sorry for my late reply.

I fully understand the issue here (thanks to your description ;) ), but unfortunately adding `update_cache=yes` to the task makes ansible run a full `apt-get update` during every run, which:
- slows down the whole run: YMMV but on a typical system it takes ~10 seconds to run a full update
- makes the run fail on the host if `apt-get update` fails a few times in a row (which can happen once in a while since the command is pretty sensitive to network flakiness)

So, instead of changing this in the present role, I'd recommend you to use the following task in all of your playbooks:

```
- name: Run "update_cache=yes" if the last one is more than 3600 seconds ago
  apt:
    update_cache: yes
    cache_valid_time: 3600
```

(extract from http://docs.ansible.com/ansible/latest/apt_module.html#examples)

This ensures that your apt cache is updated at least every hour, but not more often since that's generally unnecessary.

Setting this to `86400` seconds (i.e. a day) also sounds reasonable to me.

Let me know how this sounds to you! I'm going to go ahead and close this PR, but feel free to re-open it if you think this doesn't address the issue, thanks!

PR #81, opened 2017-10-04
Adding Agent6 (beta) support on debianoids

hush-hush:
This PR allow users to install the Agent6 (beta agent) with ansible on debianoids system.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,26 @@ Role Variables
 - `datadog_apt_key_url_new` - Override default url to the new Datadog `apt` key (in the near future the `apt` repo will have to be checked against this new key instead of the current key)
 - `datadog_agent_allow_downgrade` - Set to `yes` to allow agent downgrades on apt-based platforms (use with caution, see `defaults/main.yml` for details)
 
+Agent 6 (beta)
+--------------
+
+To upgrade or install agent6, you need to:
+
+- set `datadog_agent6` to true
+- either set `datadog_agent_version` to an existing agent6 version
+  (recommended) or leave it empty to always install the latest version.
+
+To downgrade from agent6 to agent5, you need to:
+
+- set `datadog_agent6` to false
+- pin `datadog_agent_version` to an existing agent5 version
+- set `datadog_agent_allow_downgrade` to yes
+
+Variables:
+
+- `datadog_agent6` - install an agent6 instead of agent5 (default to `false`)
+- `datadog_agent6_apt_repo` - Override default Datadog `apt` repository for agent6
+
 Dependencies
 ------------
 None
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Role Variables
 - `datadog_apt_repo` - Override default Datadog `apt` repository
 - `datadog_apt_key_url` - Override default url to Datadog `apt` key
 - `datadog_apt_key_url_new` - Override default url to the new Datadog `apt` key (in the near future the `apt` repo will have to be checked against this new key instead of the current key)
-- `datadog_allow_agent_downgrade` - Set to `yes` to allow agent downgrades on apt-based platforms (use with caution, see `defaults/main.yml` for details)
+- `datadog_agent_allow_downgrade` - Set to `yes` to allow agent downgrades on apt-based platforms (use with caution, see `defaults/main.yml` for details)
 
 Dependencies
 ------------
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -25,3 +25,29 @@ datadog_agent_version: ""
 # Set this to `yes` to allow agent downgrades on apt-based platforms.
 # Internally, this uses `apt-get`'s `--force-yes` option. Use with caution.
 datadog_agent_allow_downgrade: no
+
+########################################################################
+###            Beta-Agent6-only attributes, experimental             ###
+###   These attributes are not part of the stable API of the role    ###
+###   Subject to breaking changes between bugfix/minor versions      ###
+###                Only works on debianoids for now                  ###
+
+# Set to true to install an agent6 instead of agent5.
+#
+# To upgrade from agent5 to agent6, you need to:
+# * set 'datadog_agent6' to true, and
+# * either set 'datadog_agent_version' to an existing agent6 version
+#   (recommended) or leave it empty to always install the latest version.
+#
+# To downgrade from agent6 to agent5, you need to:
+# * 'datadog_set agent6' to false, and
+# * pin 'datadog_agent_version' to an existing agent5 version
+# * set 'datadog_agent_allow_downgrade' to yes
+datadog_agent6: no
+
+# beta APT repo where datadog-agent v6 packages are available
+datadog_agent6_apt_repo: "deb http://apt.datadoghq.com beta main"
+
+###           End of Beta-Agent6-only experimental attributes        ###
+########################################################################
+
--- a/tasks/agent6.yml
+++ b/tasks/agent6.yml
@@ -1,0 +1,44 @@
+---
+- name: Create main Datadog agant yaml configuration file (beta)
+  template:
+    src: datadog.yaml.j2
+    dest: /etc/datadog-agent/datadog.yaml
+    owner: '{{ datadog_user }}'
+    group: '{{ datadog_group }}'
+  notify: restart datadog-agent
+
+- debug: 'msg="Using deprecated `datadog_process_checks` is no longer supported in agent6, use `process` under `datadog_checks` instead"'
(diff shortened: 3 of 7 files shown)


defaults/main.yml
```diff
@@ -20,3 +20,34 @@ datadog_agent_version: ""
 # Set this to `yes` to allow agent downgrades on apt-based platforms.
 # Internally, this uses `apt-get`'s `--force-yes` option. Use with caution.
 datadog_agent_allow_downgrade: no
+
+########################################################################
+###            Beta-Agent6-only attributes, experimental             ###
+### These attributes are not part of the stable API of the cookbook  ###
```
olivielpeau: `cookbook` -> `role`

tasks/agent6.yml
```diff
@@ -0,0 +1,42 @@
...
+
+- name: Creating configuration files for extra
+  template:
+    src: config_extra.conf.j2
+    dest: /etc/datadog-agent/datadog.conf
+    owner: '{{ datadog_user }}'
+    group: '{{ datadog_group }}'
+  notify: restart datadog-agent
+
+# DEPRECATED: Remove specific handling of the process check for next major release
```
olivielpeau: let's remove this for agent6 (agent6 support can be considered a new feature, we should feel free to remove this without a major version bump :) )

defaults/main.yml
```diff
@@ -20,3 +20,34 @@ datadog_agent_version: ""
...
+# To upgrade from agent5 to agent6, you need to:
+# * set 'datadog_agent6' to true, and
+# * either set 'datadog_agent_version' to an existing agent6 version
+#   (recommended) or leave it empty to always install the latest version.
+#
+# To downgrade from agent6 to agent5, you need to:
+# * 'datadog_set agent6' to false, and
+# * pin 'datadog_agent_version' to an existing agent5 version
+# * set 'datadog_agent_allow_downgrade' to yes
+datadog_agent6: false
```
olivielpeau: nit: just for consistency with the rest of the yaml file, can we set this to `no` instead of `false`?

tasks/agent6.yml
```diff
@@ -0,0 +1,42 @@
...
+- name: Add process check configuration
+  template: src=process.yaml.j2 dest=/etc/dd-agent/conf.d/process.yaml
+  when: datadog_process_checks is defined
+  notify: restart datadog-agent
+
+- debug: 'msg="[DEPRECATION NOTICE] Using `datadog_process_checks` is deprecated, use `process` under `datadog_checks` instead"'
```
olivielpeau: let's re-word this to make it clear that `datadog_process_checks` doesn't work at all with agent 6

tasks/main.yml
```diff
@@ -5,36 +5,8 @@
...
 
-- name: Create main Datadog agent configuration file
-  template:
-    src: datadog.conf.j2
-    dest: /etc/dd-agent/datadog.conf
-    owner: '{{ datadog_user }}'
-    group: '{{ datadog_group }}'
-  notify: restart datadog-agent
+- include: agent5.yml
+  when: datadog_agent6 == false
```
olivielpeau: are you sure the comparison works as expected? (want to make sure it's actually a boolean comparison and not a string comparison: for instance, if we set `datadog_agent6: no` in the variables, does this still evaluate to the `True` boolean?)

README.md
```diff
@@ -28,7 +28,27 @@ Role Variables
...
-- `datadog_allow_agent_downgrade` - Set to `yes` to allow agent downgrades on apt-based platforms (use with caution, see `defaults/main.yml` for details)
+- `datadog_agent_allow_downgrade` - Set to `yes` to allow agent downgrades on apt-based platforms (use with caution, see `defaults/main.yml` for details)
+
+Agent 6 (beta)
+--------------
+
+To upgrade or install agent6, you need to:
+
+- set `datadog_agent6` to true
+- either set `datadog_agent_version` to an existing agent6 version
```
olivielpeau, line 39: for the Chef cookbook I chose to use a different a variable to define the agent 6 pinned version, in order to allow setting both the agent 5 and 6 version at a high level (typically at the infrastructure level) and then allow users to selectively install the agent 5 or 6 on given hosts by just flipping the `datadog_agent6` boolean to true or false.

I'm not sure the approach works as well here with Ansible, since the role doesn't handle other things that need to be updated when upgrading to Agent 6 (like the changes in the config options supported in `datadog.conf` vs `datadog.yaml`). So since the upgrade can't be made to be as simple as flipping a boolean, it might not make sense to use different version variables.

Anyway, just wanted to make sure you were aware of this, if you have any thoughts on this...

defaults/main.yml
```diff
@@ -20,3 +20,34 @@ datadog_agent_version: ""
...
+
+# beta APT repo where datadog-agent v6 packages are available
+datadog_agent6_apt_repo: "deb http://apt.datadoghq.com beta main"
+
+# Set a key to true to make the agent6 use the v2 api on that endpoint, false
+# otherwise.  Leave key value to nil to use agent6 default for that endpoint.
+# Supported keys: "series", "events", "service checks"
+datadog_agent6_use_v2_api: {}
```
olivielpeau: let's remove this variable, it's not absolutely necessary and can always be set directly on `datadog_config` if needed

templates/datadog.yaml.j2
```diff
@@ -0,0 +1,20 @@
...
+{% if datadog_config["api_key"] is not defined -%}
+api_key: {{ datadog_api_key | default('youshouldsetthis') }}
+{% endif %}
+
+{% if datadog_config["datadog_agent6_use_v2_api"] is not defined -%}
+use_v2_api: {{ datadog_agent6_use_v2_api }}
+{% endif %}
+
+{# These variables are free-style, passed through a hash -#}
+{% if datadog_config -%}
```
olivielpeau: since the format is yaml, we can probably use the filter `| to_nice_yaml` from jinja. It would ensure that we're dropping fully-valid yaml.

olivielpeau (changes requested):
Thanks, this looks 👌  overall!

I've made a few comments, most of them nits, some of them questions, let me know what you think

templates/datadog.yaml.j2
hush-hush: good catch

README.md
hush-hush: I don't think users will wildly mix agent5 and agent6. They may activate agent6 on a few specific hosts. So I don't think it's going to be a problem. We might change it later if we got feedback on it.

What do you think ?

tasks/main.yml
hush-hush: It does work, but you're right it not the best. I move condition to explicit Boolean check.

README.md
olivielpeau: hmm yeah makes sense, let's change this only if we get feedback on it 👍

tasks/agent6.yml
```diff
@@ -0,0 +1,36 @@
+---
+- name: Create main Datadog agant yaml configuration file (beta)
+  template:
+    src: datadog.yaml.j2
+    dest: /etc/datadog-agent/datadog.yaml
+    owner: '{{ datadog_user }}'
+    group: '{{ datadog_group }}'
+  notify: restart datadog-agent
+
+- name: Creating configuration files for extra
```
olivielpeau: Missed this on first review: this file won't be read by the external agents by default unfortunately

Currently we have to write a conf file for each of the external agents (trace and process), c.f. what's done in https://github.com/DataDog/chef-datadog/blob/master/recipes/_agent6_config.rb#L22-L105

A simple and dirty way to do this would be to use the same approach as the chef cookbook, i.e. simply drop these 2 conf files in their correct default locations, with the contents of a full agent5-style `datadog.conf`.

Let me know if you have any questions... It's a bit hacky but we should be able to drop it as soon as these agents can read from `datadog.yaml`

olivielpeau (changes requested):

tasks/agent6.yml
hush-hush: :+1:

templates/config_extra.conf.j2
```diff
@@ -0,0 +1,10 @@
+# Managed by Ansible
```
olivielpeau: can you remove this file altogether (and the associated task)? It doesn't really make sense to use it for agent 6

olivielpeau (changes requested):
One last comment, I'll merge and release once it's addressed, thanks!

olivielpeau (approved):

olivielpeau:
Looks good! 🚀

PR #84, opened 2017-10-26
add custom conf.d files

ismferd:
# Motivation:
I have different projects with different conf.d, and I want to use the same datadog.yaml playbook

I have put new var as custom conf.d directory, the bar must finish with a slash '/'
example:
`my_custom_confd/`

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ Role Variables
 - `datadog_apt_key_url` - Override default url to Datadog `apt` key
 - `datadog_apt_key_url_new` - Override default url to the new Datadog `apt` key (in the near future the `apt` repo will have to be checked against this new key instead of the current key)
 - `datadog_allow_agent_downgrade` - Set to `yes` to allow agent downgrades on apt-based platforms (use with caution, see `defaults/main.yml` for details)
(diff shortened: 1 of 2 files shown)


ismFerDev:
Finally, it doesn't sense for me

olivielpeau:
Hi @ismFerDev and thanks for your PR!

Could you elaborate a bit on why using the `datadog_checks` var is not a good solution to your use case?

ismferd:
yes (I will try it),

I have a single repository to handle more than one project monitored by DD so I’d like to use different conf.d files for each project, I don't want to have an unreadable with my init config.

Moreover, I have Jenkins to deploy my services, these services have different conf.d files, 
I want to  pass a var to get one or other conf.d files

ismferd:
@olivielpeau, What do you think about it?

dabcoder:
@ismferd Sorry for the long delay in updating this thread. 

> I have a single repository to handle more than one project monitored by DD so I’d like to use different conf.d files for each project...

Could you expand on what you mean by "project"? 
At this point I do not think that this is a feature we want to add to this role, so I am going to close this PR for now, thanks for your work / feedback though (and feel free to get back to us with additional context should this be an important feature you would like us to revisit).

PR #87, opened 2017-11-10
Update apt caches so you actually get the latest version of the package

lattwood:


jeffwidman:
See the problems with this approach here: https://github.com/DataDog/ansible-datadog/pull/70#issuecomment-339056391

Perhaps the PR could add both the `update_cache` and `cache_valid_time` args defaulting to 1 day...

hush-hush:
Hi @lattwood ,

Thanks for the PR, the issue was fixed in #98 and is available in the version `1.6.0`

PR #89, opened 2017-12-07
Add log example for Ansible configuration

NilsBunge:
The agent 6 now support log collection so we need to update ansible configuration example with a log section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ Example Playbooks
       tags: "mytag0, mytag1"
       log_level: INFO
       apm_enabled: "true" # has to be set as a string
+      log_enabled: true 
     datadog_config_ex:
       trace.config:
         env: dev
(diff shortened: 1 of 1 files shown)


tmichelet (approved):

README.md
```diff
@@ -106,6 +107,19 @@ Example Playbooks
           - nginx_status_url: http://example2.com:1234/nginx_status/
             tags:
               - instance:bar
+        logs:
```
olivielpeau: can you mention here (and on the `log_enabled` line above) that this is an Agent 6 feature only ? (I'm thinking this might be a bit confusing for agent 5 users otherwise)

olivielpeau (changes requested):

README.md
NilsBunge: @olivielpeau done

olivielpeau (approved):

PR #90, opened 2017-12-19
Support Agent v6 on yum-based platforms

brendanlong:
- Add the DataDog yum beta repo
 - Use `yum_repository` instead of manual files
 - Fix some places where we don't update or restart the agent
 - Create /etc/datadog

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -45,8 +45,9 @@ datadog_agent_allow_downgrade: no
 # * set 'datadog_agent_allow_downgrade' to yes
 datadog_agent6: no
 
-# beta APT repo where datadog-agent v6 packages are available
+# beta repos where datadog-agent v6 packages are available
 datadog_agent6_apt_repo: "deb http://apt.datadoghq.com beta main"
+datadog_agent6_yum_repo: "https://yum.datadoghq.com/beta/{{ ansible_userspace_architecture }}/"
 
 ###           End of Beta-Agent6-only experimental attributes        ###
 ########################################################################
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,4 +6,4 @@ install:
   - pip install ansible-lint
 
 script:
-  - ansible-lint -v $(find tests -name *yml) -x ANSIBLE0010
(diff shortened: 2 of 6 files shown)


hush-hush:
Hi @brendanlong,

Thanks for the PR ! I'll take a look at it as soon as possible (most likely next week). Thanks for your patience.

defaults/main.yml
```diff
@@ -45,8 +45,9 @@ datadog_agent_allow_downgrade: no
 # * set 'datadog_agent_allow_downgrade' to yes
 datadog_agent6: no
 
-# beta APT repo where datadog-agent v6 packages are available
+# beta repos where datadog-agent v6 packages are available
 datadog_agent6_apt_repo: "deb http://apt.datadoghq.com beta main"
+datadog_agent6_yum_repo: "https://yum.datadoghq.com/beta/{{ ansible_userspace_architecture }}/"
```
hush-hush, line 50: We don't plan on shipping 32 bits packages for the agent6. I think we could still keep the `ansible_userspace_architecture`, it will result in an error when ansible is run on unsupported architecture.

hush-hush (changes requested):
Hi @brendanlong,

Thanks a lot for this PR !
You need to add `ANSIBLE0006` to the list of excluded rules [here](https://github.com/DataDog/ansible-datadog/blob/bc66f148626ab8375b361eed610e8f368b69c081/.travis.yml#L9) for the travis test to succeed. 

Also, could you update the CHANGELOG by adding your change to a new `1.5.0` section ?

brendanlong:
@hush-hush I updated the changelog and added that new exclude rule.

defaults/main.yml
brendanlong: Do you want me to add a check for that or something?

hush-hush: I think we're good with having an error at install time. I just wanted to leave a comment here for future reference.

hush-hush:
Thanks a lot @brendanlong  for the PR !

PR #91, opened 2017-12-22
allow different checks for different host groups

snopoke:
This adds a condition to the task that creates the checks to allow deploying different checks to different host groups:
```
datadog_checks:
  ssh_check:   # gets deployed to all hosts since 'inventory_groups' is not defined
    ....
  nginx:  # only gets deployed to hosts in the 'nginx' group
    inventory_groups: 
      - nginx
    ....
```

### DIFF
--- a/tasks/agent5.yml
+++ b/tasks/agent5.yml
@@ -30,5 +30,6 @@
     dest: /etc/dd-agent/conf.d/{{ item }}.yaml
     owner: '{{ datadog_user }}'
     group: '{{ datadog_group }}'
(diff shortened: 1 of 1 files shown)


snopoke:
@jeffwidman @olivielpeau FYI

snopoke:
@jeffwidman @olivielpeau bump

robbyoconnor (commented):
I really personally don't see the point for this change -- you can just add it to the `group_vars` file and the effect will be the same. I actually do this myself. 

The other win to this is that you have everything nice and tidy.

PR #97, opened 2018-01-17
ensure apt update happens as part of install process

techstep:
During the agent installation step, apt complains that the package cannot be found. The repository for the agent is added, but `apt update` still needs to run to get the package lists. `update_cache` in the `apt_repository` tasks seems to be insufficient to get the package lists up to date.


hush-hush:
Hi @techstep,

Thanks for the PR, the issue was fixed in #98 and is available in the version `1.6.0`

PR #98, opened 2018-01-19
Update apt caches every hour to pull the latest package

hush-hush:


tasks/pkg-debian.yml
```diff
@@ -35,10 +35,14 @@
     name: datadog-agent={{ datadog_agent_version }}
     state: present
     force: "{{ datadog_agent_allow_downgrade }}"
+    update_cache: yes
+    cache_valid_time: {{ datadog_apt_cache_valid_time }}
```
olivielpeau: I'm pretty sure you need to put this between double quotes for it to work properly, i.e.: `"{{ datadog_apt_cache_valid_time }}"`

Same thing on the other occurence

olivielpeau (changes requested):
one comment about the quoting, other than that looks good to me!

olivielpeau (approved):
👌

PR #101, opened 2018-02-01
Update the role to make agent version 6 the default

hush-hush:
Update the role to make agent version 6 the default
Customers can still install agent5 as an option.

# testing

Since no package has yet been pushed to `deb http://apt.datadoghq.com/ stable 6`, this was tested with the beta repo.

# branching

A branch called `1.x` has been pushed. When agent6 goes GA, this branch will be merge to master and we will maintain the 2 branches.


README.md
```diff
@@ -67,7 +67,7 @@ Example Playbooks
     - { role: Datadog.datadog, become: yes }  # On Ansible < 1.9, use `sudo: yes` instead of `become: yes`
   vars:
     datadog_api_key: "123456"
-    datadog_agent_version: "1:5.12.3-1" # for apt-based platforms, use a `5.12.3-1` format on yum-based platforms
+    datadog_agent_version: "1:6.0.0-1" # for apt-based platforms, use a `6.0.0.1` format on yum-based platforms
```
olivielpeau: typo: for yum it should be `6.0.0-1` ;)

defaults/main.yml
```diff
@@ -12,11 +12,11 @@ datadog_user: dd-agent
 datadog_group: root
 
 # default apt repo
-datadog_apt_repo: "deb http://apt.datadoghq.com/ stable main"
+datadog_apt_repo: "deb http://apt.datadoghq.com/ stable 6"
```
olivielpeau: while we're at it (making breaking changes), can we switch the default apt repo to `https`? :)

We already install `apt-transport-https` to support https repos so it should work out-of-the-box

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 CHANGELOG
 =========
 
+# 2 / Unrelease
```
olivielpeau: typo: `Unreleased`

olivielpeau (commented):
Made a few comments, overall LGTM 👍 

Can you also check the `Role Variables` section of the README for things we should update for agent 6. If you feel like it feel free to also get rid of the deprecated options.

One last thing: you mention that we should keep supporting the 1.x branch, but I don't think we'll have to since the cost of upgrading from `1.x` to `2.0` should be minimal (it only requires setting a var to keep installing agent5)

hush-hush:
@olivielpeau thanks for the review, updated the PR and removed `datadog_process_checks`.

For the branch `1.x` I think that we will have to support it as long as agent5 is not deprecated.

truthbk (approved):
Very similar to the changes on other configuration management tools. I approve :shipit: 

Could you please some documentation (if you wish on a separate PR) for the how one would add the `apm_config` and `process_config` options to the `datadog.yaml`? In particular examples on how to enable the agents.

PR #102, opened 2018-02-05
fixed directory and filenames for checks on agent 6

dbr1993:


olivielpeau:
Thanks @dbr1993! This change is not necessary for the Agent 6 to pick up the config, it still reads yaml configs that have paths like `/etc/datadog-agent/conf.d/<integration_name>.yaml`.

Not sure yet what the default behavior of the role should be, I'll post an update once we've decided

rniedzial:
Please merge this change and publish it. While agent6 may work, it produces the config files in unexpected location which is inconsistent w/ the online documentation.

hush-hush (approved):
Thanks for the PR @dbr1993 !

olivielpeau:
@hush-hush: for upgrades of this role (from earlier versions to a version that includes this change) to work seamlessly, the role should also _remove_ any `/etc/datadog-agent/conf.d/<check_name>.yaml` file for all checks that are enabled. Otherwise, after the role upgrade, every check would end up being configured twice (once in `/etc/datadog-agent/conf.d/<check_name>.yaml`, written by the earlier version of the role, and a second time in `/etc/datadog-agent/conf.d/<check_name>.d/conf.yaml`)

hush-hush:
@olivielpeau good catch ! I'll fix this.

Thanks

hush-hush:
@olivielpeau: https://github.com/DataDog/ansible-datadog/pull/125

PR #104, opened 2018-03-05
change key to long format

dstarmanhr:
To avoid hitting this bug: https://github.com/ansible/ansible-modules-core/issues/5119

Change all keys to long format.


olivielpeau:
Hi @dstarmanhr, this issue was actually fixed with #105 , the root cause wasn't the format of the key identifier but the key itself: that key expired yesterday and Ansible fails when trying to import an expired key.

Anyway, since this is fixed now, I'm going to go ahead and close this, but thanks for opening this PR!

dstarmanhr:
@olivielpeau ah, great, thnx!

PR #114, opened 2018-03-23
Fix datadog.yaml template when used with an empty conf

### LINKED ISSUE #93: {} placed at end of /etc/datadog-agent/datadog.yaml
Using the latest version of the Ansible galaxy role from the Galaxy repo, I have the following issue. 

Replication:

Set the API variable
Enabled dd agent 6 (Believe this same error may happen regardless of agent 6)
Run the role. 

Result:

The service will fail to start and a {} is placed at the end of the datadog.yaml file. 
Removing the {} causes the service to start properly.

hush-hush:
When no option are set in `datadog_config` the generated yaml would not be valid.

This should fix: #93

### DIFF
--- a/templates/datadog.yaml.j2
+++ b/templates/datadog.yaml.j2
@@ -8,4 +8,6 @@ dd_url: {{ datadog_url | default('https://app.datadoghq.com') }}
 api_key: {{ datadog_api_key | default('youshouldsetthis') }}
 {% endif %}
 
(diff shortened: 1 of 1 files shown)


mgar:
this PR duplicates #107 😄

hush-hush:
@mgar: indeed, thanks for open #107 ! I just merged it, closing this one.

Issue #108, opened 2018-03-07
Upgrading ansible-datadog from 1.6.0 to 2.0.1 causes double APT repo entry


jharley:
Upgrading ansible-datadog from 1.6.0 to 2.0.1 on Ubuntu 16.04 (and presumably all other apt-based distributions) is causing a double entry in `/etc/apt/sources.list.d/apt_datadoghq_com.list` (one with `https` and one `http`).

This results in `apt-get update` failing due to the duplicate entry.

ofirnn:
It also makes the role broken if you already have datadog5 agent installed (apt install wouldn't work), since the path changed to /etc/datadog-agent and it doesn't exists because the duplicate entry.

jharley:
@ofirnn I was setting up the stuff to preserve the v5 agent after the APT key expiry (#103) forced me to upgrade to 2.0.1 of ansible-datadog.  Setting `datadog_agent5_apt_repo` back to the non-https value (as it was in ansible-datadog 1.6.0) fixes this issue if you keep the v5 agent

anatolebeuzon:
Same issue here. I submitted a PR to fix this: #115

excalq:
I made a similar fix to this issue, and waited until I could test it at work today. This ended up being PR #116 , but I see that you beat me to it, @oxlay. Well done.

hush-hush:
This should be fix with the new version `2.0.3`, thanks to @oxlay and @excalq for their PR !

PR #115, opened 2018-03-24
Fix bug with duplicate sources.list entry

anatolebeuzon:
# Original bug

The script on `master` branch can lead to an inconsistent state for Datadog APT repositories. The `apt_datadoghq_com.list` can look like this:

```
deb http://apt.datadoghq.com/ stable main
deb https://apt.datadoghq.com/ stable 6
```

Such a state can be reached if a previous version of the script was used (earlier than 5adf2f1) and added `deb http://apt.datadoghq.com/ stable main`. When running a more recent version of the script, it would try to remove the HTTPS repo `deb https://apt.datadoghq.com/ stable main` (which does not exist) and would add `deb https://apt.datadoghq.com/ stable 6`, thus leading to this duplicate entry.

It leads to an `apt update` error: `Duplicate sources.list entry https://apt.datadoghq.com stable Release`

This issue is mentioned in #108 

# Proposed solution

Instead of keeping track of which repositories to add/remove, we can simply declare the repository state we want to reach by using a template file for `apt_datadoghq_com.list`.

### DIFF
--- a/tasks/pkg-debian.yml
+++ b/tasks/pkg-debian.yml
@@ -11,16 +11,9 @@
   when: datadog_apt_key_url_new is defined
 
 - name: Ensure Datadog repository is up-to-date
-  apt_repository:
-    repo: "{{ datadog_apt_repo }}"
-    state: "{% if datadog_agent5 %}absent{% else %}present{% endif %}"
-    update_cache: yes
-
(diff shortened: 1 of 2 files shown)


hush-hush (approved):
Great PR, Thanks.

PR #116, opened 2018-03-26
Bugfix: Removes legacy http apt repos pre-dating #5adf2f1

excalq:
Due to the changing of the Debian APT Repos from HTTP to HTTP, without removing of stale repos which have been created in old installs of this role, this causes problems when trying to update from v5 to v6, as the stale v5 APT sources repo still exists. This adds a task to explicitly delete both v5 and v6 Datadog APT sources.list references.


hush-hush:
Thanks for the PR @excalq !

Since this is a duplicate from #115, I'm closing it.

hush-hush:
This is a better solution since it keeps the "udpate_cache: true" for apt, which is needed for fresh install.

PR #118, opened 2018-03-30
Update the example with the correct log collection variable

NilsBunge:
The variable was not up to date.


hush-hush (approved):
Thanks for the PR !

PR #122, opened 2018-04-17
Fix deprecation warning and use import_playbook over include and up min ansible version to 2.4

robbyoconnor:


robbyoconnor:
Going to close this -- apparently conditional importing doesn't work?

PR #124, opened 2018-04-23
Moises.botarro/adding fields process agent cmdline scrubbing

moisesbotarro:
Added the necessary fields to be set on the datadog.conf and datadog.yaml files to enable/disable the process args scrubbing feature on the agent 5 and 6, respectively.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,8 +60,8 @@ To enable/disable the process args scrubbing by the process agent, under the `da
 
 ```yml
 process.config:
-  scrub_args: true # default value is true
-  custom_sensitive_words: "consul_token,dd_api_key" # to expand the default list of sensitive words
+  scrub_args: true # enables scrubbing of sensitive arguments from the command line; default value is true
+  custom_sensitive_words: "consul_token,dd_api_key" # to expand the default list of sensitive words used by the cmdline scrubber
 ```
 
 Dependencies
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -70,6 +82,10 @@ Example Playbooks
       log_level: INFO
       apm_enabled: "true" # has to be set as a string
       logs_enabled: true  # log collection is available on agent 6
+      process_config: # on agent 6
```
conorbranagan: I think all of this part is for agent 6?

README.md
```diff
@@ -70,6 +82,10 @@ Example Playbooks
...
+        enabled: "true" # to enable process-agent
```
conorbranagan: Maybe say `has to be set as a string. possible values are: "true", "false" (for only container collection), "disabled" to disable the process-agent entirely`

README.md
```diff
@@ -70,6 +82,10 @@ Example Playbooks
...
+        scrub_args: true # to enable/disable process args scrubbing; default value is true
```
conorbranagan: Could be a bit clearer here, something like `enables scrubbing of sensitive arguments from the command line`

conorbranagan (commented):
Added some comments on copy, lgtm otherwise

README.md
moisesbotarro: The conf file also uses `datadog_config` to populate the template and the yaml template doesn't use the `datadog_config_ex` that's is presented on the Playbook Example. That's why I've put this comment here.

README.md
```diff
@@ -52,6 +52,18 @@ Variables:
 - `datadog_agent5` - install agent5 instead of agent6 (default to `false`)
 - `datadog_agent5_apt_repo` - Override default Datadog `apt` repository for agent5
 
+To enable the process agent on agent 5, you need to set on `datadog_config`:
```
hush-hush: Should we create a specific section in the README for the process agent ? It could make thing simpler/clearer.
What do you think ?

moisesbotarro: I agree with you. It'll make the README file clearer. I've just pushed a new commit with the suggested modification

hush-hush (approved):
Thanks @moisesbotarro

PR #128, opened 2018-04-25
Allow dry-run with our role

hush-hush:
- Fix the role to be compatible with ansible dry-run
- Fix downgrading from agent 5 to 6 on centos
- Improve our travis tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Role Variables
 - `datadog_apt_repo` - Override default Datadog `apt` repository
 - `datadog_apt_cache_valid_time` - Override the default apt cache expiration time (default 1 hour)
 - `datadog_apt_key_url_new` - Override default url to Datadog `apt` key (key ID `382E94DE` ; the deprecated `datadog_apt_key_url` variable refers to an expired key that's been removed from the role)
-- `datadog_agent_allow_downgrade` - Set to `yes` to allow agent downgrades on apt-based platforms (use with caution, see `defaults/main.yml` for details)
+- `datadog_agent_allow_downgrade` - Set to `yes` to allow agent downgrades on apt-based platforms (use with caution, see `defaults/main.yml` for details). **On centos this will only work with ansible 2.4 and up**.
 
 Agent 5 (older version)
 -----------------------
@@ -41,7 +41,7 @@ To install agent5, you need to:
 - either set `datadog_agent_version` to an existing agent5 version
   (recommended) or leave it empty to always install the latest version (`5.*`).
 
-To downgrade from agent6 to agent5, you need to:
+To downgrade from agent6 to agent5, you need to (**on centos this will only work with ansible 2.4 and up**):
 
 - set `datadog_agent5` to true
 - pin `datadog_agent_version` to an existing agent5 version
--- a/tasks/pkg-debian.yml
+++ b/tasks/pkg-debian.yml
@@ -3,6 +3,7 @@
   apt:
     name: apt-transport-https
     state: present
+  when: not ansible_check_mode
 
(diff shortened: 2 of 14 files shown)


chaosoldier:
@hush-hush 
Thanks for fixing my issue. 
When is this PR merged?

hush-hush:
@chaosoldier I'm waiting for a review from @olivielpeau. We're currently focusing on releasing the next version of the agent (6.2.0). We will merge and release the ansible role as soon as possible,

chaosoldier:
@hush-hush
Thank you for replying. 

> We will merge and release the ansible role as soon as possible,

I understood!

README.md
```diff
@@ -41,7 +41,7 @@ To install agent5, you need to:
 - either set `datadog_agent_version` to an existing agent5 version
   (recommended) or leave it empty to always install the latest version (`5.*`).
 
-To downgrade from agent6 to agent5, you need to:
+To downgrade from agent6 to agent5, you need to (**on centos this will only work with ansible 2.4+**):
```
olivielpeau: nit: `ansible 2.4 and up`?

tasks/pkg-debian.yml
```diff
@@ -1,6 +1,7 @@
 ---
 - name: Install apt-transport-https
   apt: name=apt-transport-https state=present
+  when: not ansible_check_mode
```
olivielpeau: why is this needed? Shouldn't the package be always in the apt repo?

tasks/pkg-debian.yml
```diff
@@ -18,18 +19,21 @@
   with_items:
     - "deb http://apt.datadoghq.com/ stable main"
     - "deb http://apt.datadoghq.com/ stable 6"
+  when: not ansible_check_mode
```
olivielpeau, line 30: same question

tasks/pkg-debian.yml
```diff
@@ -18,18 +19,21 @@
...
 
 - name: Ensure Datadog repository is up-to-date
   apt_repository:
     repo: "{{ datadog_apt_repo }}"
     state: "{% if datadog_agent5 %}absent{% else %}present{% endif %}"
     update_cache: yes
+  when: not ansible_check_mode
```
olivielpeau, line 37: same question

README.md
olivielpeau: you should also update the description of the `datadog_agent_allow_downgrade` variable in the `Role Variables` section of the readme

olivielpeau (commented):
One general comment: according to https://docs.ansible.com/ansible/devel/user_guide/playbooks_checkmode.html#information-about-check-mode-in-variables, the variable `ansible_check_mode` is only available on versions >=2.1 of ansible. What happens on earlier versions of ansible?

Other than that and the inline nits, LGTM! The tests are very nice 👌

hush-hush:
@olivielpeau ansible 2.1 entered end of life last year (2017-06-01). I think we can no longer support it ?

tasks/pkg-debian.yml
hush-hush: if no ansible job using apt was ever run, `--check` will fail because `python-apt` is not install.

```
fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "python-apt must be installed to use check mode. If run normally this module can auto-install it."}
```

Adding the check allow users to check their playbook without installing python-apt.

olivielpeau (approved):
LGTM!

I'm fine with dropping support of <2.1 ansible versions, but let's not forget to update the declared minimum supported version of ansible here: https://github.com/DataDog/ansible-datadog/blob/2.0.3/meta/main.yml#L6 :)

PR #129, opened 2018-04-30
include is deprecated in new versions of Ansible

Buzut:


hush-hush:
Hi @Buzut,

Thanks for opening this PR ! `include` is indeed deprecated since `2.4`, but both versions `2.2` and  `2.3` of ansible are not EOL and this role need to be compatible with them.

This means this PR need to be updated in some way to reflect that. One way to do it would be to check the version of ansible and use `include` or `include_task` depending on it. What do you think ?

Buzut:
Actually even if there's a condition on the include, the warning still appears. So better keep include till 2.2 & 2.3 reach EOL.

PR #132, opened 2018-05-14
Update platforms list

hush-hush:
This commit remove EOL ones and add new ones for EL, ubuntu and debian.

### DIFF
--- a/meta/main.yml
+++ b/meta/main.yml
@@ -7,28 +7,19 @@ galaxy_info:
   platforms:
     - name: Ubuntu
       versions:
-        - lucid
-        - maverick
-        - natty
-        - oneiric
-        - precise
-        - quantal
-        - raring
-        - saucy
         - trusty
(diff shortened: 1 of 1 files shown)


meta/main.yml
```diff
@@ -7,28 +7,18 @@ galaxy_info:
...
         - trusty
-        - utopic
-        - vivid
         - xenial
+        - artful
+        - bionic
     - name: Debian
       versions:
-        - squeeze
-        - wheezy
```
olivielpeau: IMO we should keep `wheezy` support, it'll be EOL'ed soon but is not yet. Unless there's a specific reason you want to remove it?

hush-hush: I thought it was alreadt EOL, but indeed `31st May 2018` !
Good catch, thanks !

olivielpeau (approved):

PR #134, opened 2018-05-25
SUSE/SLES support

enarciso:
* Added suse to check `os_platform` in main.yml
* recreated same method as pkg-redhat.yml with zypper


enarciso:
@hush-hush @olivielpeau Any questions or concerns regarding this?

hush-hush:
Hi @enarciso, We're currently in the middle of a new release of the agent (`6.3.0`) so we're a bit busy on it. I'll review you PR this week if everything goes well (as soon as possible in any cases).

Thank you for your patience !

enarciso:
Cool, totally understand. Thanks

tasks/pkg-suse.yml
```diff
@@ -0,0 +1,62 @@
...
+  zypper_repository:
+    name: datadog
+    description: Datadog, Inc.
+    repo: "{{ datadog_yum_repo }}"
+    enabled: yes
+    auto_import_keys: yes
+    disable_gpg_check: yes # SLES11 considers datadog's repo as untrusted
+    state: "{% if datadog_agent5 %}absent{% else %}present{% endif %}"
+
+- name: Install DataDog yum repo (agent5)
```
hush-hush: Since #136, we user `(agent5)` as a prefix. Could you update the name ?

tasks/pkg-suse.yml
```diff
@@ -0,0 +1,62 @@
...
+  zypper_repository:
+    name: datadog-beta
```
hush-hush: We should not call it `beta` since it's for the older version of the agent.

tasks/pkg-suse.yml
```diff
@@ -0,0 +1,62 @@
...
+    description: Datadog, Inc.
+    repo: "{{ datadog_agent5_yum_repo }}"
+    enabled: yes
+    auto_import_keys: yes
+    disable_gpg_check: yes # SLES11 considers datadog's repo as untrusted
+    state: "{% if datadog_agent5 %}present{% else %}absent{% endif %}"
+
+- name: Drop SysVinit script if SLES 11
```
hush-hush: We do not support sysvinit for agent6.  The agent run on Suse11 if you're running systemd.

We don't want to support any sysvinit script, especially in our ansible repo. If users want to use sysvinit they will have to ship and maintain their own script.

tasks/pkg-suse.yml
```diff
@@ -0,0 +1,62 @@
+---
+- name: Stat if RPM key already exist
```
hush-hush: Could add a condition to do this only on SLES11 and leave ansible handle the RPM key when we don't need the work around ?

tasks/pkg-suse.yml
```diff
@@ -0,0 +1,62 @@
...
+  when: not ansible_check_mode
+
+- name: Install DataDog zypper repo
+  zypper_repository:
+    name: datadog
+    description: Datadog, Inc.
+    repo: "{{ datadog_yum_repo }}"
+    enabled: yes
+    auto_import_keys: yes
+    disable_gpg_check: yes # SLES11 considers datadog's repo as untrusted
```
hush-hush: Do we you know why SLES11 considers datadog repo untrusted ? Same issue than the RPM Key ?

defaults/main.yml
```diff
@@ -20,6 +20,11 @@ datadog_yum_repo: "https://yum.datadoghq.com/stable/6/{{ ansible_userspace_archi
 datadog_yum_gpgkey: "https://yum.datadoghq.com/DATADOG_RPM_KEY.public"
 datadog_yum_gpgkey_new: "https://yum.datadoghq.com/DATADOG_RPM_KEY_E09422B3.public"
 
+# default zypper repo and keys
+datadog_zypper_repo: "https://yum.datadoghq.com/suse/stable/6/{{ ansible_userspace_architecture }}"
+datadog_zypper_gpgkey: "https://yum.datadoghq.com/DATADOG_RPM_KEY.public"
+datadog_zypper_gpgkey_sha256sum: "00d6505c33fd95b56e54e7d91ad9bfb22d2af17e5480db25cba8fee500c80c46"
```
hush-hush, line 26: This is never used anywhere, we should remove it.

hush-hush (changes requested):
Thanks a lot for this PR @enarciso !

I added a few comments or questions.

defaults/main.yml
enarciso: Hi @hush-hush,
This is my fault, I should have been using these variables. This is now fixed

enarciso:
@hush-hush,
Regarding the SysV, I went ahead and remove those 2 tasks, however that means role will fail since sles11SP4 (and lower) does not come with systemd by default (I'm not very familiar with SLES, I'm going by public vagrant images and SLES's release notes are not very clear about it).

Thanks

enarciso:
@hush-hush,
Regarding SLES11 treating Datadog's repo as untrusted, I do not have any idea. If I would guess, the service that signed the CSR is not added in SLES's as trusted

tasks/pkg-redhat.yml
```diff
@@ -21,9 +21,9 @@
     gpgkey: "{{ datadog_yum_gpgkey }}"
     state: "{% if datadog_agent5 %}absent{% else %}present{% endif %}"
 
-- name: Install DataDog yum repo (agent5)
+- name: (agent5) Install DataDog yum repo
   yum_repository:
-    name: datadog-beta
+    name: datadog
```
hush-hush: I tried this and it actually breaks agent6 install as the `state` parameter dictate, "State of the repo file".So having the same `name` here will actually remove the file created at the previous step.

Since your PR is for SUSE/SLES you could leave this file unchanged, I will rename the task and file in another PR.

tasks/pkg-suse.yml
```diff
@@ -0,0 +1,59 @@
...
+      path: /tmp/DATADOG_RPM_KEY.public
+    register: ddkey
+  - name: Download new RPM key
+    shell: "curl {{ datadog_zypper_gpgkey }} -o /tmp/DATADOG_RPM_KEY.public"
+    args:
+      warn: no
+    when: not ddkey.stat.exists
+  when: ansible_distribution_version|int == 11
+
+- name: Download new RPM key
```
hush-hush, line 14: Having the two task with the same name will break the usage of `--start-at-task` (see #136).

tasks/pkg-suse.yml
```diff
@@ -0,0 +1,59 @@
...
+    description: Datadog, Inc.
+    repo: "{{ datadog_zypper_repo }}"
+    enabled: yes
+    auto_import_keys: yes
+    disable_gpg_check: "{% if ansible_distribution_version|int == 11 %}yes{% else %}no{% endif %}" # SLES11 considers datadog's repo as untrusted
+    state: "{% if datadog_agent5 %}absent{% else %}present{% endif %}"
+
+- name: (agent5) Install DataDog yum repo
+  zypper_repository:
+    name: datadog
```
hush-hush: a state `absent` remove the repo file on centos and debian. This means that have two repository with the same `name` will remove the file created at the previous step. Could we set a unique name for each of them ?

tasks/pkg-suse.yml
```diff
@@ -0,0 +1,59 @@
...
+  when: not ansible_check_mode
+
+- name: Install DataDog zypper repo
+  zypper_repository:
+    name: datadog
+    description: Datadog, Inc.
+    repo: "{{ datadog_zypper_repo }}"
+    enabled: yes
+    auto_import_keys: yes
+    disable_gpg_check: "{% if ansible_distribution_version|int == 11 %}yes{% else %}no{% endif %}" # SLES11 considers datadog's repo as untrusted
```
hush-hush, line 34: I tried the role on suse 12, and I get an error from this. Sadly ansible is too limited to allow us to set the `repo_gpgcheck=0` option, we only have acces to `gpgcheck`. I don't know how I feel about disabling gpg check for every version of SUSE.
Did you achieve to make it work on SUSE >= 12 with the GPG check activated ?

Worst case could be to create the repository file manually with a template file.
The version on the official install page on datadog.com is:
```
[datadog]
name=Datadog, Inc.
enabled=1
baseurl=https://yum.datadoghq.com/suse/stable/6/x86_64
type=rpm-md
gpgcheck=1
repo_gpgcheck=0
gpgkey=https://yum.datadoghq.com/DATADOG_RPM_KEY.public
```

tasks/pkg-suse.yml
```diff
@@ -0,0 +1,59 @@
...
+    repo: "{{ datadog_agent5_zypper_repo }}"
+    enabled: yes
+    auto_import_keys: yes
+    disable_gpg_check: "{% if ansible_distribution_version|int == 11 %}yes{% else %}no{% endif %}" # SLES11 considers datadog's repo as untrusted
+    state: "{% if datadog_agent5 %}present{% else %}absent{% endif %}"
+
+- name: Install pinned datadog-agent package
+  zypper:
+    name: "datadog-agent={{ datadog_agent_version }}"
+    state: present
```
hush-hush, line 50: When downgrading from agent6 to agent5 zypper will refuse to install the agent since the version is lower.

We need to set the `oldpackage` option:
```yaml
oldpackage: "{{ datadog_agent_allow_downgrade }}"
```

hush-hush (changes requested):
Thank you for your quick update !

I tested the role on suse 12, almost worked. I think we're almost there.

I also ask @gmmeyer for a review as he as more experience than me on SUSE.

tasks/pkg-suse.yml
gmmeyer: Zypper has two different types of gpg checks, the repo one and the other one. If you're importing the gpg key properly, it should work for packages. (I don't know ansible well enough to know if t's being imported properly.

gmmeyer:
@hush-hush I'm happy with this, it looks fine to me from the suse standpoint!

enarciso:
Thanks for the feedback @gmmeyer! Let me clean it up further based of @hush-hush comments and give you guys an update. 

Thanks

tasks/pkg-suse.yml
enarciso: This is fixed, see 9201723

tasks/pkg-suse.yml
enarciso: This is fixed using the `force` param in the zypper module.

tasks/pkg-suse.yml
enarciso: @hush-hush and @gmmeyer, Zypper not importing GPG keys properly may be the same issue filed with OpenSuse, see https://github.com/openSUSE/zypper/issues/144

Let me know if you are okay with gpgcheck disabled with SLES11 OS.

Thanks

tasks/pkg-suse.yml
enarciso: After further testing, it looks like SLES11's Zypper ignores the `force` param. In order to downgrade an agent from 6 to 5, the agent needs to be uninstalled first before installing 5. I can either make this note in the README.md file or create a tasks to uninstall current agent first with `datadog_agent_allow_downgrade` conditions then install v5. Let me know what works for you guys.

Thanks

enarciso:
Hi @hush-hush and @gmmeyer, let me know if there are further issues that I need to fix. So far this is working for me on both SLES11 & 12.

Thank you very much!

hush-hush (approved):

hush-hush:
Thank you for this awesome PR @enarciso !

Sorry it took so long to merge, I'll publish a new version of the role today.

enarciso:
Woo! Glad this is merged. I'll switch to this soon instead of my fork. Thanks again gents!

Issue #135, opened 2018-06-04
Avoid using same task name


xp-1000:
Hello,

I would create a generic task because it could probably avoid some limitations that I don't see yet and could be a good practice but in fact I have a very specific problem on only one task.

So, the object of this issue is to rename all duplicated tasks between agent5 & agent6 to prevent confusion and corner case as you already did on : 

- for agent6  : `Create main Datadog agent yaml configuration file (beta)`
- for agent5 : `Create main Datadog agent configuration file`

In my case, the problem is on : `Create a configuration file for each Datadog check` which is the same on agent6 or agent5.

Indeed, I use this command to configure my agent : 

```
ansible-playbook -i localhost, -c local --start-at-task='Create a configuration file for each Datadog check' agent.yml
```

But it seems that ansible don't handle double occurence of a same name of a task (at least with --start-at-task parameter).

So, if you can rename it to another name (i.e. suffix (beta) or (agent6) in name), it could allow me to workaround on this limitation.

In the same time, I will open an issue on ansible repo to handle the start at task with on multiple same task name 

thanks

hush-hush:
Hi @xp-1000,

Thank you for opening this issue. This should indeed be fix. Sadly our backlog is a bit full right now. Even if it breaks the usage of `--start-at-task` it's not a critical bug.

I will create a ticket in our backlog and we will try to fix this as soon as possible. In the meantime, if you would like to open a PR I would be more than happy to merge it ! If you do open a PR, I think it would be better to prefix every agent5 task with `[agent5]` and leave agent6 task as it is right now since it's the default.

Thanks !

xp-1000:
Hi @hush-hush, thanks for your quick answer.
I understand and I agree : it is not a critical bug. 
I will open a PR soon so following your advice.
Have a good day.

xp-1000:
@hush-hush, here is the PR : https://github.com/DataDog/ansible-datadog/pull/136
while `[agent5]` seems not supported, instead I used `(agent5)` as prefix.
I tried to also harmonize name and order in addition to only renaming task.
Feel free to make suggstions, I am available to do changes as you want.

hush-hush:
Version `2.2.0` with the PR has been release

xp-1000:
Thank you very much for your quick actions, have a good day !

Edit: just for information I created an issue on ansible repo about this need
https://github.com/ansible/ansible/issues/41122

PR #136, opened 2018-06-05
Harmonize tasks names between agent5 and agent6

xp-1000:
This PR aims to : 
- harmonize tasks name between v5 and v6 agent
- avoid using same name for multiple different tasks (to prevent breaking `--start-at-task` option) #135

### DIFF
--- a/tasks/agent6.yml
+++ b/tasks/agent6.yml
@@ -1,10 +1,10 @@
 ---
-- name: Create /etc/datadog-agent
+- name: Create Datadog agent config directory
   file:
     dest: /etc/datadog-agent
     state: directory
 
-- name: Create main Datadog agent yaml configuration file (beta)
+- name: Create main Datadog agent configuration file
   template:
     src: datadog.yaml.j2
     dest: /etc/datadog-agent/datadog.yaml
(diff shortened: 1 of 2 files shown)


tasks/agent6.yml
```diff
@@ -1,17 +1,31 @@
...
-- name: Create main Datadog agent yaml configuration file (beta)
+- name: Create main Datadog agent configuration file
   template:
     src: datadog.yaml.j2
     dest: /etc/datadog-agent/datadog.yaml
     owner: "{{ datadog_user }}"
     group: "{{ datadog_group }}"
   notify: restart datadog-agent
 
+- name: Ensure datadog-agent is running
```
hush-hush: Shouldn't we first create every configuration files before starting the agent, or am I missing something ?

hush-hush (changes requested):
Thanks a lot for the PR @xp-1000 !
It looks good to go, just a small detail

tasks/agent6.yml
xp-1000: hmmm let met test that ^^

xp-1000:
I reverted my change on the task order @hush-hush let me know if you would like other changes.

tasks/agent6.yml
xp-1000: it works fine but I reverted to original order as this does not bring any profit and forces agent to (re)start 2 times instead only 1 at first install. let met know if you need other changes

hush-hush (approved):
Thanks a lot for this PR @xp-1000 !

I'll release the new version today.

PR #144, opened 2018-07-04
Avoid restarts from yum install workaround for resolved issue 20608

kcd83:
Would you please consider merging this as the bug was fixed a few versions ago. The workaround triggered datadog-agent to restart on every run and so I have fixed that too.

Thanks

### DIFF
--- a/tasks/pkg-redhat.yml
+++ b/tasks/pkg-redhat.yml
@@ -39,17 +39,6 @@
   when: (datadog_agent_version != "") and (not ansible_check_mode)
   notify: restart datadog-agent
 
-# Using `yum` directly to work around an Ansible bug:
-# https://github.com/ansible/ansible/issues/20608
-- name: Install latest datadog-agent package issue 20608 workaround
-  command: yum install -y datadog-agent
-  register: yum_install_result
-  changed_when: ( yum_install_result.stdout_lines | last ) != "Nothing to do"
-  when:
-  - datadog_agent_version == ""
-  - ansible_version is version_compare(2.3, '<')
-  notify: restart datadog-agent
-
 - name: Install latest datadog-agent package
   yum:
     name: datadog-agent
@@ -57,5 +46,4 @@
     state: latest
   when:
   - datadog_agent_version == ""
-  - ansible_version is version_compare(2.3, '>=')
   notify: restart datadog-agent
--- a/tasks/pkg-redhat.yml
+++ b/tasks/pkg-redhat.yml
@@ -43,6 +43,8 @@
 # https://github.com/ansible/ansible/issues/20608
 - name: Install latest datadog-agent package issue 20608 workaround
   command: yum install -y datadog-agent
+  register: yum_install_result
+  changed_when: ( yum_install_result.stdout_lines | last ) != "Nothing to do"
   when:
   - datadog_agent_version == ""
   - ansible_version is version_compare(2.3, '<')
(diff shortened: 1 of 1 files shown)


kcd83:
#147

(Playing with how github references PRs from issues)

kcd83:
@olivielpeau could you please take a look at this?

tasks/pkg-redhat.yml
```diff
@@ -41,7 +41,21 @@
...
 # https://github.com/ansible/ansible/issues/20608
-- name: Install latest datadog-agent package
+- name: Install latest datadog-agent package issue 20608 workaround
   command: yum install -y datadog-agent
-  when: datadog_agent_version == ""
+  register: yum_install_result
+  changed_when: ( yum_install_result.stdout_lines | last ) != "Nothing to do"
+  when:
+  - datadog_agent_version == ""
+  - ansible_version is version_compare(2.3, '<')
```
dabcoder: We actually only support >= 2.4: https://github.com/DataDog/ansible-datadog/blob/ec5e5fd736982e87d4a31f06a8ba484f97ab8340/meta/main.yml#L6 (and version 2.3 has been marked as EOL: https://docs.ansible.com/ansible/2.4/release_and_maintenance.html#release-status) so we can remove that case and keep the one below given that the `#20608` issue is now resolved in >= 2.4.

tasks/pkg-redhat.yml
```diff
@@ -41,7 +41,21 @@
...
   notify: restart datadog-agent
+
+- name: Install latest datadog-agent package
+  yum:
+    name: datadog-agent
+    update_cache: yes
+    state: latest
+  when:
+  - datadog_agent_version == ""
+  - ansible_version is version_compare(2.3, '>=')
```
dabcoder: Given that this role support >= 2.4 we can drop this condition.

dabcoder (changes requested):
Left a couple of notes, thanks!

dabcoder:
@kcd83 Did you have a chance to review my comments above?

tasks/pkg-redhat.yml
kcd83: Good idea, much simpler

tasks/pkg-redhat.yml
```diff
@@ -39,9 +39,11 @@
...
-# https://github.com/ansible/ansible/issues/20608
 - name: Install latest datadog-agent package
-  command: yum install -y datadog-agent
-  when: datadog_agent_version == ""
+  yum:
+    name: datadog-agent
+    update_cache: yes
+    state: latest
+  when:
+  - datadog_agent_version == ""
```
kcd83, line 48: This could got back to one line, and do you want `not ansible_check_mode` to be consistent with above

kcd83:
Good points, yes I finally have a moment to check this

tasks/pkg-redhat.yml
dabcoder: Good point, we do it for Debian based distros: https://github.com/DataDog/ansible-datadog/blob/34e4275e079d71b3ea775ad2ac3d73e2361dd0cc/tasks/pkg-debian.yml#L61 but not for Red Hat nor Suse. We could add for both. I believe your branch does not include the `pkg-suse.yaml` file so you'll need to rebase on master to make that change (if you want?).

Thanks!

dabcoder:
Took care of that in https://github.com/DataDog/ansible-datadog/pull/199, but thanks for your work here @kcd83, much appreciated!

kcd83:
Thanks @dabcoder !

Issue #43, opened 2017-01-31
Removing entries from `datadog_checks` doesn't delete the file


bgerstle:
Resulting in the datadog agent attempting to run checks which are likely fail. 

### Steps to reproduce
1. Configure playbook w/ datadog role & an arbitrary check
2. Run playbook
3. Remove entry from `datadog_checks` and replace it with a different one
4. Run playbook again

### Expected result
Check is removed and agent is restarted, only running the new check.

### Actual result
Old check is still run, as the `/etc/dd-agent/checks.d/<check>.yaml` file is still present.

Ideally we'd be doing immutable deploys and this wouldn't be an issue, but c'est la vie.

olivielpeau:
Hi @bgerstle,
Thanks for your feedback! Indeed once a check config is not listed in `datadog_checks` the role  stops keeping track of the config file, so it doesn't delete it.

There are a few different things that could be done by the role to address the case you described:
- have a variable (named `datadog_disabled_checks` for instance) that would allow listing checks for which the role would make sure no configuration file is present in `/etc/dd-agent/checks.d/`. This would still allow users to have check configs that are not managed by the role (as long as they're not listed in the variable, the role wouldn't touch them).
- a simpler but less flexible approach would be to have a flag variable (named `datadog_disable_untracked_checks` for instance) that would make the role remove all the check configs in `/etc/dd-agent/checks.d/` that are not tracked.

Any thoughts?

jharley:
I'm into the idea of something like `datadog_dsiable_untracked_checks` for hosts that aren't immutable.

bgerstle:
@olivielpeau sorry for the (very) delayed response! Those options sound usable, but I'm curious what you think of either:

1. Following the apparent Ansible pattern of using a `state` variable which allows users to specify `state: absent`
2. Removing all checks except those specified for the given host (including default checks like ntp & such)

jeffwidman:
I prefer the flexibility of being able to have checks that we manage outside of Ansible. For example, sometimes when I'm developing, I end up needing to sanity test checks in production and I don't want those checks removed if someone else unbeknownst to me runs the Ansible play.

bgerstle:
[...]

PR #151, opened 2018-08-22
Remove datadog checks

Jno21:
This PR is a proposal for the issue [43](https://github.com/DataDog/ansible-datadog/issues/43).

By setting: `datadog_disable_untracked_checks` to `true` all the files that are not in `datadog_checks` and `datadog_additional_checks` will be removed.

`datadog_additional_checks` allow you to add some checks that won't be remove while running the ansible role with `datadog_disable_untracked_checks: true`.

If you have any feedback it will be appreciated, I am still new to Ansible :)


rromanchuk:
thanks @Jno21, much needed, especially in the beginning of integration when the velocity of changes to these checks are high from testing/optimizing/rearrangingu integrations. Been burned a couple of times not realizing old checks were still running

dabcoder:
@Jno21 Sorry about the long delay here. I'm currently testing this and it looks pretty good so far. Your branch is now out of sync with master, would you mind rebasing it on master then (we've split the agent 6 yaml file in 2 - Linux and Windows, and removed support for agent 5 on Windows)

Jno21:
Hi, Sorry about the delay also. I rebased to be in sync with master, I also did it for Windows (I couldn't test it tho, so I am not 100% sure about it).
If the windows one is not correct I can drop the commit and we can still merge.

tasks/agent5-linux.yml
```diff
@@ -26,6 +26,21 @@
     enabled: no
   when: not datadog_skip_running_check and not datadog_enabled
 
+- name: Register all checks directories present in datadog
+  find:
+    paths: /etc/datadog-agent/conf.d/
```
dabcoder: With agent 5 it's `/etc/dd-agent/` not `/etc/datadog-agent/`.

tasks/agent5-linux.yml
```diff
@@ -26,6 +26,21 @@
...
+    pattern: "*.yaml"
+    file_type: directory
+  register: datadog_conf_directories
+  when: datadog_disable_untracked_checks
+
+- name: Delete checks not present in datadog_tracked_checks
+  file:
+    path: "/etc/datadog-agent/conf.d/{{ item }}.yaml"
```
dabcoder: Same as above.

dabcoder (changes requested):
Left a couple of notes, thanks for getting back to us!

tasks/agent5-linux.yml
Jno21: Thx, I just fixed it

tasks/agent5-linux.yml
Jno21: Thx, I just fixed it

dabcoder (approved):

izeau:
Is there any advancement on this? I'd love to be able to use this role without relying on @Jno21's fork! Thanks.

tasks/agent5-linux.yml
```diff
@@ -26,6 +26,21 @@
     enabled: no
   when: not datadog_skip_running_check and not datadog_enabled
 
+- name: Register all checks directories present in datadog
+  find:
+    paths: /etc/dd-agent/conf.d/
+    pattern: "*.yaml"
+    file_type: directory
```
KSerrania: On Agent 5, this directly matches files (as there are no `<check>.d` folders), so the file type should be changed accordingly (otherwise Ansible doesn't match anything since there are no directories ending in `.yaml`)
```suggestion
    file_type: file
```

tasks/agent5-linux.yml
```diff
@@ -26,6 +26,21 @@
...
+  register: datadog_conf_directories
+  when: datadog_disable_untracked_checks
+
+- name: Delete checks not present in datadog_tracked_checks
+  file:
+    path: "/etc/dd-agent/conf.d/{{ item }}.yaml"
```
KSerrania, line 47: This will remove any user-defined configuration of the checks, but in the case of core checks (disk, memory, etc.), they will still be running with the default configuration, as the default config file (`{{ item }}.yaml.default`) will still be present.
Is this expected? Or should the default files also be removed?

(Same question for Agent 6, with the `conf.d/{{ item }}.d/conf.yaml.default` files)

README.md
```diff
@@ -29,6 +29,8 @@ Role Variables
 - `datadog_checks` - YAML configuration for agent checks to drop into:
   + `/etc/datadog-agent/conf.d/<check_name>.d/conf.yaml` for agent6
   + `/etc/dd-agent/conf.d` for agent5.
+- `datadog_disable_untracked_checks` - Set to `true` to remove all checks not present in `datadog_checks` and `datadog_additional_checks`.
+- `datadog_additional_checks`: List of additional checks that won't be remove if `datadog_disable_untracked_checks` is set to `true`.
```
KSerrania:
```suggestion
- `datadog_additional_checks`: List of additional checks that won't be removed if `datadog_disable_untracked_checks` is set to `true`.
```

KSerrania (commented):
Thanks for the PR! I left some questions and suggestions.

There's a merge conflict because the README got updated on master (the list of attributes has become a table), could you update your PR to resolve the conflict?

tasks/agent5-linux.yml
lesteenman: While it would be useful to keep them as a base case, I think it'll make things a lot complexer when people start modifying the defaults - and there are likely use cases where you want to remove the defaults anyway.

Maybe an additional config option (`datadog_disable_default_checks`)?

lesteenman:
@Jno21, do you plan to finish this PR? I might be able to pick it up otherwise.

README.md
Jno21: Done

tasks/agent5-linux.yml
Jno21: Hey ! I followed @lesteenman idea by adding a `datadog_disable_default_checks` option (true or false). So it will remove all defaults checks or not.
If you want to go further with the idea you can pick it up to choose which one you will prefer to remove (list).

Jno21:
Hey ! Sorry for the long wait, I made corrections like asked and added the idea of @lesteenman ! 
I didn't have time to test it (I am in vacation). So if somebody could make sure it is correct and working and after we are good to go :) 
It will be cool to merge this PR :D

README.md
```diff
@@ -36,6 +36,9 @@ To deploy the Datadog Agent on hosts, add the Datadog role and your API key to y
...
 | `datadog_agent_major_version`              | The major version of the Agent to install. The possible values are 5, 6, or 7 (default). If `datadog_agent_version` is set, it takes precedence otherwise the latest version of the specified major is installed. Setting `datadog_agent_major_version` is not needed if `datadog_agent_version` is used. |
 | `datadog_checks`                           | YAML configuration for Agent checks to drop into: <br> - `/etc/datadog-agent/conf.d/<check_name>.d/conf.yaml` for Agent v6 and v7, <br> - `/etc/dd-agent/conf.d` for Agent v5.                                                                                                                            |
+| `datadog_disable_untracked_checks` | Set to `true` to remove all checks not present in `datadog_checks` and `datadog_additional_checks`. |
+| `datadog_additional_checks` | List of additional checks that won't be remove if `datadog_disable_untracked_checks` is set to `true`. |
```
kayayarai:
```suggestion
| `datadog_additional_checks` | List of additional checks that won't be removed if `datadog_disable_untracked_checks` is set to `true`. |
```

kayayarai (approved):
small grammar fix, otherwise good from a documentation perspective.

albertvaka:
Thanks for your PR and sorry for the (very long) wait 🙇 LGTM and merging! 👍

PR #155, opened 2018-10-05
stop generating datadog v6 conf when api key is not defined.

guitarrapc:
I have some situation which want to manage configuration without ansible, and there need some switch to trigger config generation.

It seems like rather adding new property but existing api key check is much reasonable.

### DIFF
--- a/tasks/agent6.yml
+++ b/tasks/agent6.yml
@@ -11,6 +11,7 @@
     owner: "{{ datadog_user }}"
     group: "{{ datadog_group }}"
   notify: restart datadog-agent
+  when: datadog_api_key is defined
 
 - name: Ensure configuration directories are present for each Datadog check
   file:
@@ -43,6 +44,7 @@
     owner: "{{ datadog_user }}"
     group: "{{ datadog_group }}"
   notify: restart datadog-agent
+  when: datadog_api_key is defined
 
 - name: Create process agent configuration file
   template:
@@ -51,6 +53,7 @@
     owner: "{{ datadog_user }}"
     group: "{{ datadog_group }}"
   notify: restart datadog-agent
+  when: datadog_api_key is defined
 
 - name: Ensure datadog-agent is running
   service:


hush-hush:
hI @guitarrapc,

Thanks for the PR ! May I ask why are you using the datadog role if you don't wan't any configuration file ? Is your target to only install the agent ?

guitarrapc:
No, and yes.
My most environments are generating configuration file with this role, but eventually some of them are not managed by Ansible.
So the motivation is "I want manage whole environments datadog installation and configuration with ansible-datadog, but can manage config generation is much prefer."

enarciso:
Hello, apologies to interject. Would it not be simpler to create a child group (in ansible inventory) that should not create a configuration and the group_var will not have an entry in the `datadog_checks:` key?

example:
```bash
○ → tree -a
.
└── inventories
    ├── group_vars
    │   └── noconfig.yml
    └── hosts

2 directories, 2 files

○ → cat inventories/hosts 
[host1]
192.168.1.10

[host2]
192.168.1.11

[hosta]
192.168.2.10

[hostb]
192.168.2.11

[noconfig:children]
host1
hosta

[group_of_managed_config:children]
host2
hostb

○ → cat inventories/group_vars/noconfig.yml 
datadog_api_key: 1234test1234
datadog_agent_allow_downgrade: yes
datadog_agent_version: "1:6.4.2-1"
datadog_config:
  log_level: INFO
  hostname: "{{ ansible_hostname }}"
datadog_checks:
```
The sample above only manages the `/etc/datadog/datadog.yaml`.
But if you're implying to simply install the agent and completely not manage the configuration including the `datadog.yaml` main config. Then perhaps you can just create a different tasks in the playbook


```bash
---
- hosts: group_of_managed_config
  gather_facts: true
  become: true
  roles:
    - datadog

- hosts: noconfig
  tasks:
    - name: install agent
       apt:
           name: datadog-agent
```

hush-hush:
Hi @guitarrapc,

I think @enarciso suggestion is the way to go as it leverage ansible features without adding complexity to this role. Thanks a lot @enarciso for it !

I'm going to close this PR, but please feel free to reopen it if needed.

PR #156, opened 2018-10-19
Add support for "site" option to send data to Europe

hush-hush:


README.md
```diff
@@ -19,6 +19,12 @@ ansible-galaxy install Datadog.datadog
 Role Variables
 --------------
 
+- `datadog_site` - The site of the Datadog intake to send Agent data to.
+  Defaults to 'datadoghq.com', set to 'datadoghq.eu' to send data to the EU
+  site. This option is only available with agent version >= 6.6.0.
+- `datadog_url` - The host of the Datadog intake server to send Agent data to,
```
olivielpeau: I don't think we should expose the `url` as a "top-level" role variable. Users should very rarely have to set this variable, and if they need to they can use `datadog_config` like other agent config option.

README.md
```diff
@@ -19,6 +19,12 @@ ansible-galaxy install Datadog.datadog
 Role Variables
 --------------
 
+- `datadog_site` - The site of the Datadog intake to send Agent data to.
```
olivielpeau, line 23: nit: this should probably be documented below `datadog_api_key`, since the api key remains the only mandatory variable to set, whereas `datadog_site` is optional

README.md
```diff
@@ -131,10 +137,12 @@ This example will configure the PostgeSQL check through **autodiscovery**:
 
 Example Playbooks
 -----------------
+
+Sending data to Datadog US (default) and configuringa few checks.
```
olivielpeau: typo: `configuring a`

templates/datadog.yaml.j2
```diff
@@ -1,7 +1,14 @@
 # Managed by Ansible
 
-{% if datadog_config["dd_url"] is not defined -%}
-dd_url: {{ datadog_url | default('https://app.datadoghq.com') }}
+{% if datadog_config["dd_url"] is not defined
```
olivielpeau: I don't think we need to handle this logic here, we can let the Agent handle the options' precedence

templates/datadog.yaml.j2
```diff
@@ -1,7 +1,14 @@
...
+   and datadog_url is not defined
+   and datadog_config["site"] is not defined -%}
+site: {{ datadog_site | default('datadoghq.com') }}
```
olivielpeau: I'd let the Agent define the default value as well, i.e. if the ansible variable is not set, I'm in favor of not setting the option at all in this config.

olivielpeau (commented):
Thanks!

Made a few comments, overall towards pushing as much logic as possible to the Agent so that the role's logic remains as simple as possible.

(overall this is how the `site` option was designed: handle as much the logic as possible on the Agent side so that installation/configuration methods remain as simple as possible)

README.md
hush-hush: Done

olivielpeau (approved):
💯

Issue #146, opened 2018-07-20
Agent not reporting logs due to dd-agent missing systemd-journal group


endorama:
Hello, we are using this role for deploying `datadog` agents on our instances.

While configuring `journald` log collection on 2 different EC2 instances, I found a strange behaviour: *instance-2* was correctly reporting logs, *instance-1* was not. The same playbook has been applied to both instances, using this role.

The datadog configuration applied is:
```yaml
datadog_api_key: <omitted>
datadog_config:
  logs_enabled: true
datadog_checks:
  journald:
    logs:
    - type: journald
      include_units:
        - our-custom.service
```

Both agent are starting without errors. *instance-1* agent logs report tailing from journald as expected:
```
instance-1 systemd[1]: Started "Datadog Agent".
.
.
.
... INFO | (logs.go:29 in Start) | Starting logs-agent
... INFO | (tailer.go:45 in Start) | Start tailing journal  default
.
.
.
```

After some *hours* (:persevere:) of debugging, reading and re reading the docs I found a subtle difference between the two instances:
- *instance-2*
    ```
    # id dd-agent
    uid=114(dd-agent) gid=117(dd-agent) groups=117(dd-agent),101(systemd-journal)
    ```
- *instance-1*
    ```
    # id dd-agent
    uid=114(dd-agent) gid=117(dd-agent) groups=117(dd-agent)
    ```

My bad as I probably add it manually only on one of the two, and as your [installation documentation](https://docs.datadoghq.com/integrations/journald/#installation) for journald log collection states:
> Add the dd-agent user to the systemd-journal group by running:

What I actually didn't expected was:
- this step to be out of this ansible role
- no documentation of this **mandatory** step in the readme ( or on the online docs ); I also suspect that other integration would require other groups to be assigned to the `dd-agent` user
- the logs showing "Start tailing journal  default", without any kind of error about the impossibility to tail journald

I don't know what would be the best way to inform users of this role about this, but would be nice if you could do something about as it has been a really difficult and long debugging session.

Thank you :slightly_smiling_face:

mgu:
what could be great is a way to add dd-agent user to extra groups (`systemd-journal`,  `adm`, ...) in this role

NilsBunge:
Hello @endorama and @mgu ,

Thanks for reaching out and for your suggestions. 

By default, the agent tries to connect to the journal and reads what he can. In your case it was nothing but as it could read the file, it did not raise an error in the status.
[...]

PR #165, opened 2018-11-20
Adding dd-agent user to systemd-journal when logs is enabled

hush-hush:
This is needed by the logs agent to collect logs from systemd.

Fixes: #146.

### DIFF
--- a/tasks/agent6.yml
+++ b/tasks/agent6.yml
@@ -52,6 +52,25 @@
     group: "{{ datadog_group }}"
   notify: restart datadog-agent
 
+- block: # Adding dd-agent to the systemd-journal if logs collection is enabled
+  - name: Testing if systemd-journal group exists
+    # using cat to discard error status from grep when systemd-journal is not
+    # found. Using 'ignor_errors' still print the error message and disabling
+    # logs output for this task still logs a message in red saying that logs
+    # were discarded.
+    shell: "grep -E '^systemd-journal:' /etc/group | cat"
+    args:
+      warn: no
+    register: systemd_group_exists
+  - name: Add the dd-agent user to the systemd-journal for logs collection
+    user:
+      name: dd-agent
+      groups: systemd-journal
+      append: yes
+    notify: restart datadog-agent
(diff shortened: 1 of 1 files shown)


tasks/agent6.yml
```diff
@@ -52,6 +52,25 @@
...
+      warn: no
+    register: systemd_group_exists
+  - name: Add the dd-agent user to the systemd-journal for logs collection
+    user:
+      name: dd-agent
+      groups: systemd-journal
+      append: yes
+    notify: restart datadog-agent
+    when: systemd_group_exists.stdout|length > 0
+  when: datadog_enabled and not ansible_check_mode and datadog_config.get("logs_enabled") == true
```
okunc, line 72: this will actually add dd-agent to systemd-journal every time you set logs_enabled=true, but in some cases you want to listen for 'type: file' and not 'type: journald'. This can be a bit tricky to check but I believe that this condition should include an additional check, something like if datadog_checks/*/logs/type=journald, then it would be bulletproof :]

tasks/agent6.yml
```diff
@@ -52,6 +52,25 @@
...
+    # found. Using 'ignor_errors' still print the error message and disabling
+    # logs output for this task still logs a message in red saying that logs
+    # were discarded.
+    shell: "grep -E '^systemd-journal:' /etc/group | cat"
+    args:
+      warn: no
+    register: systemd_group_exists
+  - name: Add the dd-agent user to the systemd-journal for logs collection
+    user:
+      name: dd-agent
```
okunc, line 67: don't hardcode! ;]
```suggestion
      name: "{{ datadog_user }}"
```

tasks/agent6.yml
hush-hush: Good catch, I'm not sure how a such a test could be simply done. I will double check with the different deployment target with the logs-team and see if this is worth it.

Based on you PR #167 the simplest might be to add 'journald' directly to the list when needed.

hush-hush:
Closing this PR for now until #167 is merged

PR #166, opened 2018-11-20
Add the agent to the docker group if the docker check is enabled

hush-hush:
This is mandatory to monitor Docker.

Fix: #130

PR #160, opened 2018-10-28
Group membership for datadog user

fbentouati:
Hello!

By default the Datadog agent is part of root group. For my need, I want to specify a particular group to the Datadog user.

My need is I want to forward my logs to the Datadogs plateforme, for that the Datadog agent need to have the good right on theses files.

Thanks!

PR #167, opened 2018-11-28
add datadog_additional_groups and datadog_use_local_repo switches

okunc:
This is to make DD deploy more flexible - two changes [switches] are proposed:

1] `datadog_use_local_repo` - Set to `true` only in case you have local repo with datadog sources. When set to true, repo is not handled at all - this assumes you have your e.g. satellite server already configured and containing the datadog rpm. This switch applies to yum repo management only.

2] `datadog_additional_groups` - List of additional groups for `datadog_user`.

comment: this may seem to interfere with these two [and I like them!] ..
- https://github.com/DataDog/ansible-datadog/pull/165
- https://github.com/DataDog/ansible-datadog/pull/166

.. but my proposal brings extra option to add DD user to other groups which may not be detected or which are simply out of scope of the auto-detection. E.g. you may need cassandra group to read debug logs etc.. In a way this could be also used to [test-only / edge cases / not recommended] add DD user to root group.

There's also another interfering pull request https://github.com/DataDog/ansible-datadog/pull/160 - but that one allows only one group, while my proposal has a list.

thanks a lot for coop / update and have a nice day all!
Ondrej

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,8 @@ Role Variables
 - `datadog_apt_cache_valid_time` - Override the default apt cache expiration time (default 1 hour)
 - `datadog_apt_key_url_new` - Override default url to Datadog `apt` key (key ID `382E94DE` ; the deprecated `datadog_apt_key_url` variable refers to an expired key that's been removed from the role)
 - `datadog_agent_allow_downgrade` - Set to `yes` to allow agent downgrades on apt-based platforms (use with caution, see `defaults/main.yml` for details). **On centos this will only work with ansible 2.4 and up**.
+- `datadog_additional_groups` - List of additional groups for `datadog_user`.
+- `datadog_use_local_repo` - Set to `true` only in case you have local repo with datadog sources. When set to true, repo is not handled at all - this assumes you have your e.g. satellite server already configured and containing the datadog rpm. This switch applies to yum repo management only.
 
 Agent 5 (older version)
 -----------------------
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -13,3 +13,9 @@
 
 - include: agent6.yml
   when: not datadog_agent5
+
+- name: add "{{ datadog_user }}" user to additional groups
+  user: name="{{ datadog_user }}" groups="{{ item }}" append=yes
+  with_items: "{{ datadog_additional_groups }}"
+  when: datadog_additional_groups is defined and datadog_additional_groups != ''
+  notify: restart datadog-agent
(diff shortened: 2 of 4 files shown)


okunc:
Oops I messed up the commit message - all the referenced pull requests relate to my point 2.

README.md
```diff
@@ -36,6 +36,8 @@ Role Variables
 - `datadog_apt_cache_valid_time` - Override the default apt cache expiration time (default 1 hour)
 - `datadog_apt_key_url_new` - Override default url to Datadog `apt` key (key ID `382E94DE` ; the deprecated `datadog_apt_key_url` variable refers to an expired key that's been removed from the role)
 - `datadog_agent_allow_downgrade` - Set to `yes` to allow agent downgrades on apt-based platforms (use with caution, see `defaults/main.yml` for details). **On centos this will only work with ansible 2.4 and up**.
+- `datadog_additional_groups` - List of additional groups for `datadog_user`.
+- `datadog_use_local_repo` - Set to `true` only in case you have local repo with datadog sources. When set to true, repo is not handled at all - this assumes you have your e.g. satellite server already configured and containing the datadog rpm. This switch applies to yum repo management only.
```
hush-hush, line 40: Why not use the `datadog_yum_repo` variable to set your own repo ?

tasks/main.yml
```diff
@@ -13,3 +13,9 @@
 
 - include: agent6.yml
   when: not datadog_agent5
+
+- name: add "{{ datadog_user }}" user to additional groups
+  user: name="{{ datadog_user }}" groups="{{ item }}" append=yes
+  with_items: "{{ datadog_additional_groups }}"
```
hush-hush, line 19: `groups` can take a list directly, I don't think we need the `with_items`.

hush-hush (changes requested):
Thanks a lot for this PR @okunc !

The user/group parts seems like a very good idea.

For the `datadog_use_local_repo` I'm not sure I see the advantage compare to using the `*_repo` variables in default/main.yml.

Could you either split this PR in 2 or remove the part about `datadog_use_local_repo`, that way I can merge the group part !

dabcoder:
@okunc Any feedback regarding the comments we made last year?

dabcoder:
Closing in favor of https://github.com/DataDog/ansible-datadog/pull/194 but thanks for the proposal and original work there.

okunc:
> Closing in favor of #194 but thanks for the proposal and original work there.

@dabcoder sorry for late reply and thanks a lot for making this happen!

PR #175, opened 2018-12-19
Fix configuration items order

jpiron:
dict doesn't ensure order leading to configuration to change
between Ansible playbook runs.


hush-hush (approved):
Thanks for this PR @jpiron !

Issue #161, opened 2018-10-30
Question: Is it possible to Install but not restart DD?


tim-w-photo:
We are building immutable infrastructure with Ansible via Packer.

We build an AMI once, and then scale it and deploy it across our environments (and set its hostname at deploy time).
So, when we build the AMI, we want to install DataDog, but not start it until we next bring the machine up.

**Is it possible to install DataDog (and its startup scripts) with this role, but not actually restart it?**

At the moment, we get lots of up and down messages in our DataDog slack channel for "localhost" whilst packer is building our AMIs.

tim-w-photo:
Adding 
```
datadog_enabled: no
```
to `meta/main.yml` solved it for me. I thought it might not start on next boot, but it did.

PR #177, opened 2019-01-30
Document datadog_enabled role variable

valscion:
Hi there! It seems that there exists a `datadog_enabled` variable that can be set to `false` to disable the `datadog-agent` from starting.

However, that role variable was not documented, so I wasn't sure whether it's OK to use it. Our use case is that we'd want development boxes to not start the agent, but still run all the other datadog installation steps to get the development boxes as close to the production systems as possible.

It seems like there are other issues / pull requests also referring to this role variable based on a quick search:

- https://github.com/DataDog/ansible-datadog/pull/153
- https://github.com/DataDog/ansible-datadog/issues/161

Cheers! Let me know if there's any questions :relaxed:


hush-hush (approved):
Thanks for the PR @valscion.

The reason it was not documented: `datadog_enabled` was used internally for testing but it seems to be useful for others.

valscion:
Thanks! This indeed helps us prebuild a machine but not have datadog enabled during the prebuild phase, but only at deployment time :relaxed:

PR #179, opened 2019-03-01
Add support for managing datadog agent for windows via ansible

derekwbrown:
Provides support for managing the Datadog Agent for windows via ansible.


tasks/main.yml
```diff
@@ -9,7 +9,18 @@
...
 
 - include: agent5.yml
-  when: datadog_agent5
+  when: datadog_agent5 and ansible_os_family != "Windows"
 
 - include: agent6.yml
-  when: not datadog_agent5
+  when: not datadog_agent5 and ansible_os_family != "Windows"
+
+- fail:
```
hush-hush: I would move this to `tasks/pkg-windows.yml` so it's self contained

tasks/win_agent_version.yml
```diff
@@ -0,0 +1,4 @@
+---
+- name: set agent download filename to a specific version
+  set_fact:
+    dd_download_url: "https://s3.amazonaws.com/ddagent-windows-stable/ddagent-cli-{{ datadog_agent_version }}.msi"
```
hush-hush: Do you think moving the URL to the `defaults/main.yml` so customers can change it make sens ? Do we have users that cache our msi ?

tasks/main.yml
```diff
@@ -9,7 +9,18 @@
   when: ansible_os_family == "Suse"
 
 - include: agent5.yml
-  when: datadog_agent5
+  when: datadog_agent5 and ansible_os_family != "Windows"
```
hush-hush, line 15: Could we add the `-linux` suffix to `agent5.yaml` and `agent6.yaml` files ?

tasks/main.yml
```diff
@@ -9,7 +9,18 @@
...
 
 - include: agent6.yml
-  when: not datadog_agent5
+  when: not datadog_agent5 and ansible_os_family != "Windows"
+
+- fail:
+    msg: "The Datadog ansible role does not currently support Agent 5"
+  when: datadog_agent5 and ansible_os_family == "Windows"
+
+- include: pkg-windows.yml
```
hush-hush: nit-pick: could we move this at the beginning of the file with the other `pkg-*.yml` task ?

tasks/agent6-win.yml
```diff
@@ -0,0 +1,52 @@
+---
+- name: Create main Datadog agent configuration file
```
hush-hush, line 2: Do we care about the owner of those files ? Aren't they going to belong to `administrator` ?

hush-hush (changes requested):

tasks/win_agent_version.yml
derekwbrown: Good question. I thought about it, as it would help me for testing, but it didn't seem generic enough for customer use... but I'm willing to be convinced :). Note that in the specific version case it's computed.

tasks/agent6-win.yml
derekwbrown: They should inherit the rights from the directory, which the installer creates.

tasks/main.yml
derekwbrown: I was going for minimal change.

tasks/agent6-win.yml
hush-hush: Good to know !

tasks/main.yml
hush-hush: I think a rename is fine and make things more explicit

tasks/win_agent_version.yml
hush-hush: I think it doesn't cost us anything to move the `https://s3.amazonaws.com/ddagent-windows-stable/` part of the URL to the default (for both latest and a specific version). worst case it doesn't change anything, best case it gives user more flexibility.

hush-hush (approved):

PR #180, opened 2019-03-01
Switch from include to include_tasks to speed up the role

rouge8:
This patch switches from Ansible's `include` to `include_tasks`. With `include`, the file is always included and the `when:` conditional is evaluated on each task. With `include_tasks`, the conditional is evaluated once.

In my tests with 36 hosts, this cuts the time spent in the Datadog role in half!

### DIFF
--- a/meta/main.yml
+++ b/meta/main.yml
@@ -3,7 +3,7 @@ galaxy_info:
   author: 'Brian Akins, Dustin Brown & Datadog'
   description: Install Datadog agent and configure checks
   license: Apache2
-  min_ansible_version: 2.2
+  min_ansible_version: 2.4
   platforms:
     - name: Ubuntu
       versions:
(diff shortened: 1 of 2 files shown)


rouge8:
@hush-hush any interest in this? When I tested this versus 2.5.0, It cut the time to provision 36 hosts from 180s to 90s. I imagine 2.6.0 is even slower than 2.5.0 because it has more `include:`s 😩

hush-hush:
Hi @rouge8,

Sorry for the very late answer, I was busy on other project. I'll take a look this week. Moving to `include_tasks` seems like an obvious good idea, I just have to check the oldest version of Ansible we support have them (that's why we didn't used it last time I checked).

hush-hush (approved):
Thanks a lot for this PR @rouge8 !

PR #185, opened 2019-03-15
fix type datadog-agent to dd-agent for datadog-agent5

charsyam:
I found some datadog-agent folders should be dd-agent. because it is for datadog-agent5


hush-hush (approved):
Thanks for the fix @charsyam !

PR #173, opened 2018-12-05
Changes get_url.sha256sum to get_url.checksum

histamineblkr:
The get_url.sha256sum has been deprecated in favor of the get_url.checksum parameter. Corresponding Ansible documentation: https://docs.ansible.com/ansible/2.7/modules/get_url_module.html

PR #188, opened 2019-04-03
Fix ansible lint issues

bhundven:
Some of these changes actually fix ansible style issues, others tell ansible-lint to ignore the issues.

I didn't fix the remaining `[303] yum used in place of yum module`, as that is fixed in #173. (good job @histamineblkr)

### DIFF
--- a/tasks/pkg-suse.yml
+++ b/tasks/pkg-suse.yml
@@ -34,7 +34,7 @@
   register: datadog_zypper_repo
 
 # refresh zypper repos only if the template changed
-- name: refresh Datadog zypper_repos
+- name: refresh Datadog zypper_repos # noqa 503
   command: zypper refresh datadog
   when: datadog_zypper_repo.changed and not ansible_check_mode
   args:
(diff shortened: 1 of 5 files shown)


bhundven:
ansible-lint
https://docs.ansible.com/ansible-lint/index.html

Maybe also look at pre-commit:
https://pre-commit.com/

tasks/pkg-suse.yml
```diff
@@ -1,16 +1,16 @@
 ---
-- block: # Work around due to SNI check for SLES11
```
albertvaka, line 2: Have you tested this change doesn't break SLES11?

bhundven: I will test it tonight.

dabcoder: @bhundven did you have a chance to run your test(s)?

bhundven: I have not. I don't have a suse setup to test on. Do you have any suggestions for testing this?

albertvaka: If you split the PR in two, we can merge everything else now and wait for this part until someone can test it.

dabcoder: @bhundven any feedback on Albert's last comment? Let us know if you don't have time to split this PR and we can take care of the Debian part for now...

dabcoder:
Followed up in #200, but thanks for initiating the work @bhundven

PR #186, opened 2019-03-20
Releasing 3.0.0

hush-hush:

PR #194, opened 2019-05-13
Adds dd-user to optional groups

dabcoder:
Tested on Ubuntu and CentOS.

This picks up the work done in https://github.com/DataDog/ansible-datadog/pull/167 to add the `dd-user` to other groups (to be specified by the user). Linux only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ Role Variables
 - `datadog_agent_allow_downgrade` - Set to `yes` to allow agent downgrades on apt-based platforms (use with caution, see `defaults/main.yml` for details). **On centos this will only work with ansible 2.4 and up**.
 - `use_apt_backup_keyserver` - Set `true` to use the backup keyserver instead of the default one
 - `datadog_enabled` - Set to `false` to prevent `datadog-agent` service from starting. Defaults to `true`
+- `datadog_additional_groups` - Comma separated list of additional groups for the `datadog_user`.
 
 Agent 5 (older version)
 -----------------------
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Role Variables
 - `datadog_agent_allow_downgrade` - Set to `yes` to allow agent downgrades on apt-based platforms (use with caution, see `defaults/main.yml` for details). **On centos this will only work with ansible 2.4 and up**.
 - `use_apt_backup_keyserver` - Set `true` to use the backup keyserver instead of the default one
 - `datadog_enabled` - Set to `false` to prevent `datadog-agent` service from starting. Defaults to `true`
-- `datadog_additional_groups` - Comma separated list of additional groups for the `datadog_user`.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -39,6 +39,7 @@ Role Variables
 - `datadog_agent_allow_downgrade` - Set to `yes` to allow agent downgrades on apt-based platforms (use with caution, see `defaults/main.yml` for details). **On centos this will only work with ansible 2.4 and up**.
 - `use_apt_backup_keyserver` - Set `true` to use the backup keyserver instead of the default one
 - `datadog_enabled` - Set to `false` to prevent `datadog-agent` service from starting. Defaults to `true`
+- `datadog_additional_groups` - Comma separated list of additional groups for the `datadog_user`.
```
hush-hush: Let's specify that it's a linux only feature (and not windows user groups)

tasks/main.yml
```diff
@@ -19,3 +19,9 @@
 
 - include_tasks: agent6-win.yml
   when: not datadog_agent5 and ansible_os_family == "Windows"
+
+- name: add "{{ datadog_user }}" user to additional groups - Linux
```
hush-hush: why not add this to agent6-linux.yaml instead ?

hush-hush (changes requested):

hush-hush (commented):
Also let's add a line to the changelog

tasks/main.yml
dabcoder: Yes, if we want to provide this for agent 6 only, works for me.

dabcoder:
> Also let's add a line to the changelog

I think we can edit the changelog in one go once we finish the 3.0.0 release? https://github.com/DataDog/ansible-datadog/pull/186

hush-hush (approved):

endorama:
Hello @hush-hush, is this feature the proposed solution to #165?

Thank you

dabcoder:
@endorama This feature allows to add the `dd-agent` users to other groups, but does not automatically add it to the `systemd-journal` group if log collection is enabled. You will, however, be able to do so by adding `systemd-journal` to the `datadog_additional_groups` optional key in your playbook.
I hope that clarifies the situation.

endorama:
@dabcoder yes, but I just now noticed I linked to the wrong issue.  

What I wanted to ask was if this should be considered the solution to #146 or if there are plans in the future to implement automatic addition to `systemd-journal` group for `dd-agent` user. 
Sorry for the confusion.

In any case this is surely helpful in mitigating #146 for the time being, so thank you for the feature!

dabcoder:
@endorama as the `journald` integration is not enabled by default, we are not going to add the `dd-agent` user to `systemd-journal`, so I believe using `datadog_additional_groups` should be the solution for this use case indeed.

endorama:
Thanks for the clarification! I'll report this in #146 for clarity.

PR #200, opened 2019-06-03
Ansible lint for the agent version

dabcoder:
Picks up the work started in https://github.com/DataDog/ansible-datadog/pull/188/ but for ansible-lint issues only (which recommends to avoid comparing to empty string).

### DIFF
--- a/meta/main.yml
+++ b/meta/main.yml
@@ -30,6 +30,6 @@ galaxy_info:
         - 2012
         - 2012R2
         - 2016
-  categories:
+  galaxy_tags:
     - monitoring
 dependencies: []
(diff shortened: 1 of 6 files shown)


meta/main.yml
```diff
@@ -30,6 +30,6 @@ galaxy_info:
         - 2012
         - 2012R2
         - 2016
-  categories:
+  galaxy_tags:
```
albertvaka, line 33: Is this compatible with the oldest ansible version we support?

albertvaka (approved):

meta/main.yml
dabcoder: 2.5, yes: https://docs.ansible.com/ansible/2.5/reference_appendices/galaxy.html?highlight=galaxy_tags#import-a-role as far as I can see

albertvaka: 👍

albertvaka (approved):

PR #201, opened 2019-06-07
sha256sum: has been deprecated

dabcoder:
Picks up the work done in https://github.com/DataDog/ansible-datadog/pull/173


albertvaka (approved):
Looks good to me as long as it doesn't break the older ansible versions we support.

dabcoder:
Yes indeed, it's had this option since 2.0 and since we support >= 2.5... https://docs.ansible.com/ansible/2.5/modules/get_url_module.html

albertvaka:
👍

PR #203, opened 2019-07-22
[suse] Import the new GPG key

KSerrania:
In the near future, we're going to sign our RPM packages with our
newer GPG key. The RHEL task already imports the new key, but not the SUSE
task, so the new key import step has been added to it.


defaults/main.yml
```diff
@@ -31,6 +31,8 @@ datadog_yum_gpgkey_new: "https://yum.datadoghq.com/DATADOG_RPM_KEY_E09422B3.publ
 datadog_zypper_repo: "https://yum.datadoghq.com/suse/stable/6/{{ ansible_userspace_architecture }}"
 datadog_zypper_gpgkey: "https://yum.datadoghq.com/DATADOG_RPM_KEY.public"
 datadog_zypper_gpgkey_sha256sum: "00d6505c33fd95b56e54e7d91ad9bfb22d2af17e5480db25cba8fee500c80c46"
+datadog_zypper_gpgkey_new: "https://yum.datadoghq.com/DATADOG_RPM_KEY_E09422B3.public"
```
hush-hush: Could we rename the variable for something without `new` like `datadog_zypper_gpgkey_E09422B3`  (in order to preprare for the next key rotation) ?

hush-hush (changes requested):
Could you also update the CHANGELOG and the README to add a bit of documentation ?

hush-hush (approved):

Issue #202, opened 2019-07-16
Feature request: add support for post-install custom tasks


yiwork:
Currently inside the datadog ansible role, there's no way to add custom post install tasks without having to fork the repo:

Suggestion, add a post_task folder to include all other tasks that might need to be invoked after main configuration:

in `datadog-datadog` role main task, modify to be the following
```
- include: agent6.yml
  when: not datadog_agent5

- include: post_tasks/*.yml
```

Make certain that post_task is git-ignored so we can keep pulling updates without needing to fork the repo, and still have the role perform other tasks post config/install. 

Perfect use-case:
https://github.com/DataDog/ansible-datadog/issues/130

Not everyone needs docker integration, but if you do, all customers need to do is to create a new task definition in `post_tasks` folder to add datadog-agent user to the docker group. 

I could see as well a `pre_tasks` folder being useful as well, to prepare for datadog role invocation.

dabcoder:
Thanks for the detailed information.

> Currently inside the datadog ansible role, there's no way to add custom post install tasks without having to fork the repo:

True, but shouldn't these `post_tasks` and `pre_tasks` be handled at the playbook level and not at the role level? Or am I missing something there?

By the way [unrelated], for #130, we've added an option to add the `dd-agent` user to other user defined groups.

yiwork:
`True, but shouldn't these post_tasks and pre_tasks be handled at the playbook level and not at the role level? Or am I missing something there?`

Yup, I am asking that we borrow an idea from playbook and using it in a role. This will make datadog role a lot more flexible for your customers during deployment, without having to remember to always include certain tasks in playbooks. 

For example - I'm leveraging datadog socket file to send datadog metrics. For every single playbook, I always have to remember(aside from adding datadog to docker group), to create the socket file. 

Another example - for some legacy servers that I have (that I cannot just rebuild from scratch), I have to conduct some cleaning up prior to invoking the role. 

Given how Datadog is increasingly adding new features (which is great), and are centralizing the deployment of all those products upon the deployment and customization of just one agent. I think this would make your role a lot easier to use and update.

yiwork:
FWIW - this feature would help us solve https://github.com/DataDog/ansible-datadog/issues/43 without having you guys get involved. 

So is this one - https://github.com/DataDog/ansible-datadog/issues/73 we could easily taken care of this one ourselves by adding a simple task in the `post_tasks` folder, and again, key being without having getting you guys involved. 

[...]

PR #204, opened 2019-08-02
Add pre and post_tasks

dabcoder:
Attempts to address #202.

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,6 +268,11 @@ datadog_config_ex:
     custom_sensitive_words: "consul_token,dd_api_key"
 ```
 
+Additional tasks
+----------------
+
+`pre_tasks` and `post_tasks` folders allow to run user defined tasks. `pre_tasks` for tasks to be executed before executing any tasks defined in the `tasks` folder. `post_tasks` for those to be executed after the ones defined in the `tasks` folder.
+
 Known Issues and Workarounds
 ----------------------------
 
--- a/README.md
+++ b/README.md
@@ -271,7 +271,7 @@ datadog_config_ex:
 Additional tasks
 ----------------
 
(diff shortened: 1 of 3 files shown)


tasks/main.yml
```diff
@@ -19,3 +19,9 @@
 
 - include_tasks: agent6-win.yml
   when: not datadog_agent5 and ansible_os_family == "Windows"
+
+- include_tasks: post_tasks/*.yml
```
hush-hush, line 23: Should we ignore errors in post/pre tasks ?

We should also document this in the README

hush-hush (changes requested):

tasks/main.yml
dabcoder: Good point. I would say we shoudn't, since it wouldn't give users any hints as to where the issue(s) lie should they come across any while adding pre and post tasks (and running their playbooks)?

README.md
```diff
@@ -268,6 +268,11 @@ datadog_config_ex:
     custom_sensitive_words: "consul_token,dd_api_key"
 ```
 
+Additional tasks
+----------------
+
+`pre_tasks` and `post_tasks` folders allow to run user defined tasks. `pre_tasks` for tasks to be executed before executing any tasks defined in the `tasks` folder. `post_tasks` for those to be executed after the ones defined in the `tasks` folder.
```
hush-hush: `any tasks defined in the `tasks` folder` I don't think users are aware of the layout of our role.
I would rephrase to something like:

`pre_tasks` and `post_tasks` folders allow to run user defined tasks. `pre_tasks` for tasks to be executed before executing any tasks from the Datadog role and `post_tasks` for those to be executed after.

hush-hush (approved):

PR #207, opened 2019-08-14
Remove agent v5 INI config templates from v6 tasks

### LINKED ISSUE #206: The trace-agent.conf and process-agent.conf files aren't needed in agent6 tasks
The agent6-linux.yml and agent6-win.yml tasks still install both the trace-agent.conf and process-agent.conf files, even though they are not used (afaict) in v6, and in fact will be improperly formed when generated from the datadog.conf.j2 template which now uses YAML for the datadog_config (and which gets rendered incorrectly as Python objects in the template, when using YAML nested dictionaries, etc.)  These unused files create confusion because ansible shows them as updating with clearly bogus results when changing/adding-to the datadog_config YAML variable.

The .conf install tasks should be removed from tasks/agent6-linux.yml, possibly with some deprecation message for those who don't realize they need to migrate their config over to datadog_config.

chadnetzer:
The v6 agent doesn't need these v5 config files.  Further, the new
datadog_config variable is YAML, and the datadog.conf.j2 template
renders the YAML content incorrectly (as Python datatypes), producing
confusing and invalid ansible updates to these unused files.

Fixes #206

### DIFF
--- a/tasks/agent6-linux.yml
+++ b/tasks/agent6-linux.yml
@@ -42,22 +42,6 @@
   with_items: "{{ datadog_checks|list }}"
   notify: restart datadog-agent
 
-- name: Create trace agent configuration file
-  template:
-    src: datadog.conf.j2
-    dest: /etc/datadog-agent/trace-agent.conf
-    owner: "{{ datadog_user }}"
-    group: "{{ datadog_group }}"
-  notify: restart datadog-agent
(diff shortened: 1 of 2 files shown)


KSerrania:
Superseded by #272 which got merged, I'm closing this PR and the underlying issue.

PR #208, opened 2019-08-27
File/directory mode

urusha:
Set file 0440 and dir 0550 mode, since it's world readable, user writable, and may contain private data like passwords, api keys, etc. Windows is affected too, we could use win_acl module, but this needs tests.


alopezz:
Closing as it's no longer relevant.

PR #210, opened 2019-10-03
Add support for creating system-probe configuration

kevinconaway:
This PR adds support for writing the `system-probe.yaml` configuration file to the ansible playbook.

If the system probe is enabled, the corresponding service will be started.

@DataDog/burrito 
@hush-hush

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,6 +217,8 @@ Sending data to Datadog US (default) and configuring a few checks.
   roles:
     - { role: Datadog.datadog, become: yes }
   vars:
+    system_probe_config:
+      enabled: true
     datadog_api_key: "123456"
     datadog_agent_version: "1:6.8.0-1" # for apt-based platforms, use a `6.8.0-1` format on yum-based platforms
     datadog_config:
@@ -317,6 +319,10 @@ or "disabled" (to disable the Process Agent entirely)
 - `scrub_args` - enables the scrubbing of sensitive arguments from a process command line. Default value is true
 - `custom_sensitive_words` - expands the default list of sensitive words used by the cmdline scrubber
 
+#### System Probe
+
+The network performance monitoring system probe is configured under the `system_probe_config` variable.  Any variables nested underneath will be written to the `system-probe.yaml`
+
 #### Example of configuration
 ```yml
 datadog_config:
@@ -324,6 +330,9 @@ datadog_config:
     enabled: "true" # has to be set as a string
     scrub_args: true
     custom_sensitive_words: ['consul_token','dd_api_key']
+system_probe_config:
+  enabled: true
+  sysprobe_socket: /opt/datadog-agent/run/sysprobe.sock
 ```
 
 ### Agent 5
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ Supports most Debian and RHEL-based Linux distributions, and Windows.
   * [APM](#apm)
   * [Process Agent](#process-agent)
       - [Variables](#variables)
+      - [System Probe](#system-probe)
       - [Example of configuration](#example-of-configuration)
     + [Agent 5](#agent-5)
       - [Example of configuration](#example-of-configuration-1)
@@ -217,8 +218,6 @@ Sending data to Datadog US (default) and configuring a few checks.
   roles:
     - { role: Datadog.datadog, become: yes }
   vars:
-    system_probe_config:
-      enabled: true
     datadog_api_key: "123456"
     datadog_agent_version: "1:6.8.0-1" # for apt-based platforms, use a `6.8.0-1` format on yum-based platforms
     datadog_config:
@@ -277,6 +276,8 @@ Sending data to Datadog US (default) and configuring a few checks.
         version: 1.11.0
       datadog-postgres:
         action: remove
+    system_probe_config:
+      enabled: true
 ```
 
 Example for sending data to EU site:
@@ -321,7 +322,9 @@ or "disabled" (to disable the Process Agent entirely)
 
 #### System Probe
 
-The network performance monitoring system probe is configured under the `system_probe_config` variable.  Any variables nested underneath will be written to the `system-probe.yaml`
+The [network performance monitoring](https://docs.datadoghq.com/network_performance_monitoring/) system probe is configured under the `system_probe_config` variable.  Any variables nested underneath will be written to the `system-probe.yaml`.
+
+Currently, the system probe only works on Linux with the Agent 6 version and beyond.
 
 #### Example of configuration
 ```yml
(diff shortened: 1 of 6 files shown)


tasks/agent6-linux.yml
```diff
@@ -10,6 +10,13 @@
     dest: /etc/datadog-agent
     state: directory
 
+- name: Create system-probe configuration file
+  template:
+    src: system-probe.yaml.j2
+    dest: /etc/datadog-agent/system-probe.yaml
+    owner: "root"
```
leeavital: should this be owned by dd-agent ? The process-agent needs to read the system-probe.yaml file, so making the owner root could be a problem

kevinconaway: Its currently owned by root in our environments.

sunhay: I do think it should be owned by root - is there a way to confirm that it'll be readable by dd-agent? (or additional config to change the chmod settings?)

README.md
```diff
@@ -217,6 +217,8 @@ Sending data to Datadog US (default) and configuring a few checks.
   roles:
     - { role: Datadog.datadog, become: yes }
   vars:
+    system_probe_config:
```
hush-hush: Please move this section after the `datadog_config` one. We try to have the most used option first.
Moving it line 274 or 279 would be better.

README.md
```diff
@@ -317,13 +319,20 @@ or "disabled" (to disable the Process Agent entirely)
 - `scrub_args` - enables the scrubbing of sensitive arguments from a process command line. Default value is true
 - `custom_sensitive_words` - expands the default list of sensitive words used by the cmdline scrubber
 
+#### System Probe
```
hush-hush: Please update the TOC of the readme too

README.md
```diff
@@ -317,13 +319,20 @@ or "disabled" (to disable the Process Agent entirely)
...
+
+The network performance monitoring system probe is configured under the `system_probe_config` variable.  Any variables nested underneath will be written to the `system-probe.yaml`
```
hush-hush: Could you add a link to the official system-probe doc page. Also please mention that this is only supported for Agent6 Linux.

tasks/agent6-linux.yml
```diff
@@ -10,6 +10,13 @@
     dest: /etc/datadog-agent
     state: directory
 
+- name: Create system-probe configuration file
```
hush-hush: Is the system prob only available one linux ? Do we plan to have support for Windows at some point ?

Please move this after line 52 (before or after `Create trace agent configuration file`). We want to create the datadog.yaml first (in case it fails the error raise to the customer will be about the main agent).

tasks/agent6-linux.yml
hush-hush: I was told it will be own by root as a security requirement. If it's own by dd-agent then we must move it inside the datadog.yaml like every other "datadog" configuration.

hush-hush: Is there a situation where a user might want to change this ? Should we make it configurable ?

hush-hush (changes requested):

tasks/agent6-linux.yml
kevinconaway: > Is the system prob only available one linux ? 

Yes

> Do we plan to have support for Windows at some point ?

I believe so.

tasks/agent6-linux.yml
kevinconaway: Not to my knowledge but I suppose if someone is using linux capabilities, they may create a different user for this purpose.  It can't hurt to template it out.

kevinconaway:
@sunhay Per our convo, I've updated this PR to also start the probe if it is enabled.

tasks/agent6-linux.yml
hush-hush: Well, templating it means we support if officially, I let you choose if it's something we want to do.

tasks/agent6-linux.yml
```diff
@@ -58,14 +58,29 @@
...
+
 - name: Ensure datadog-agent is running
   service:
     name: datadog-agent
     state: started
     enabled: yes
   when: not datadog_skip_running_check and datadog_enabled and not ansible_check_mode
 
-- name: Ensure datadog-agent, datadog-agent-process and datadog-agent-trace are not running
+- name: Ensure datadog-agent-sysprobe is running if enabled
```
hush-hush: why are we checking twice that the system-probl is is running ? Since it's added to the list line 93

hush-hush (changes requested):

tasks/agent6-linux.yml
```diff
@@ -58,14 +58,29 @@
...
+  service:
+    name: datadog-agent-sysprobe
+    state: started
+    enabled: yes
+  when: not datadog_skip_running_check and datadog_enabled and not ansible_check_mode and system_probe_config is defined and system_probe_config["enabled"]
+
+- name: Ensure datadog-agent, datadog-agent-process, datadog-agent-trace and datadog-agent-sysprobe are not running
```
sunhay: Will this still work on older agent versions that don't have the system probe bundled?

tasks/agent6-linux.yml
kevinconaway: @sunhay Do we want this templated?

tasks/agent6-linux.yml
hush-hush: Indeed, I don't think it will

kevinconaway: Good catch.  I've updated this to use the [service_facts](https://docs.ansible.com/ansible/latest/modules/service_facts_module.html#service-facts-module) module to determine if the service is present.

tasks/agent6-linux.yml
kevinconaway: Removed

sunhay (approved):
LGTM

We should wait on @hush-hush's review as well before merging :)

kevinconaway:
Per our convo yesterday, I am going to update the perms to be `0640` and the group owner to be the agent user.  I am testing out internally first before doing so.

hush-hush (approved):

tasks/agent6-linux.yml
stefanux: > I was told it will be own by root as a security requirement. If it's own by dd-agent then we must move it inside the datadog.yaml like every other "datadog" configuration.

what is the security-gain of a root-ownership especially on this file? if your thinking is: "the unpriviledged datadog-user should not be able to modify its own configuration" than *all* config should be owned by root and only readable by the datadog_group.

sunhay: The system-probe needs to run with elevated privileges as it uses [eBPF](https://lwn.net/Articles/740157/). 

We do not want the unprivileged datadog-user to make changes to a configuration file that a privileged binary reads - though since the unprivileged user still needs to access it, it should have read-only rights.

PR #220, opened 2019-11-05
Agent 7 support

KSerrania:
### What does this PR do?

Adds Agent 7 install support.

The `datadog_agent5` option is deprecated and replaced by `datadog_agent_major_version`, which can be set to 5, 6 or 7.
If set, the `datadog_agent_version` option must have the major version specified in `datadog_agent_major_version`.

For Agent 6 and 7, the `datadog_apt_repo`, `datadog_yum_repo`, `datadog_zypper_repo` options have been changed: if they're not set, the module will use the official Datadog repository
for the major version specified in `datadog_agent_major_version`. If they're set, the specified repository will be used, regardless of the `datadog_agent_major_version` set.

### WIP

- ~Test Agent 7 install~ (tested with `7.15.0~beta.1`), and various upgrade / downgrade scenarios. 
- ~Manual test on SUSE~
- Manual test on Windows
- ~Add version changes to README + upgrade notes~


albertvaka:
README needs to be updated

defaults/main.yml
```diff
@@ -23,30 +23,53 @@ datadog_agent_binary_path_windows: "C:\\Program Files\\Datadog\\Datadog Agent\\b
...
 
-# default apt repo
-datadog_apt_repo: "deb https://apt.datadoghq.com/ stable 6"
+# Major version of the Agent that will be installed.
+# Possible values: 5, 6, 7
+datadog_agent_major_version: 7
+
+# Pin agent to a version. Highly recommended.
+# Defaults to the latest version of the major version chosen in datadog_agent_major_version
+# The version pinned needs to have the same major as datadog_agent_major_version
```
hush-hush: I don't think we should have to set `datadog_agent_major_version` if we set `datadog_agent_version`. This duplicate the same information.

Usage should be:
- set nothing → install latest version from 7
- set `datadog_agent_major_version` → install the latest version of this major
- set `datadog_agent_version` → install that specific version (ignoring `datadog_agent_major_version` if also set, with maybe a warning message ?)

defaults/main.yml
```diff
@@ -23,30 +23,53 @@ datadog_agent_binary_path_windows: "C:\\Program Files\\Datadog\\Datadog Agent\\b
...
+datadog_agent_version: ""
+
+# default apt repo and keyserver
+
+# By default, will use the official apt Datadog repository for the chosen major version
+# Agent 5: deb https://apt.datadoghq.com/ stable main
+# Agent 6: deb https://apt.datadoghq.com/ stable 6
+# Agent 7: deb https://apt.datadoghq.com/ stable 7
+# Use this option to override the repository used
+datadog_apt_repo: ""
```
hush-hush, line 42: Having the default in `defaults/main.yml` is better that hardcoding them in the tasks I think. This file is the first place where people look for that kind of information and it avoid the possibility of using an uninitialized var.

tasks/pkg-debian.yml
```diff
@@ -31,21 +31,36 @@
...
     update_cache: yes
-  when: not ansible_check_mode
+  when: (datadog_apt_repo | length == 0) and (datadog_agent_major_version == 6) and (not ansible_check_mode)
+
+- name: (Agent 5) Ensure Datadog repository is up-to-date
+  apt_repository:
+    repo: "deb https://apt.datadoghq.com stable main"
+    state: present
+    update_cache: yes
+  when: (datadog_apt_repo | length == 0) and (datadog_agent_major_version == 5) and (not ansible_check_mode)
```
hush-hush: This actually would leave the previous repo (either 6 or 7) when switching major. This means that when switching from 7 to 6 without pining a version we would still install the latest of 7.

what about something more generic like this (haven't tested it):

```
￼- name: (Agent {{ item.key }} ) Ensure Datadog repository is up-to-date
￼  apt_repository:
￼    repo: "{{ item.value }}"
￼    state: "{% if item.key ==  datadog_agent_major_version and  datadog_apt_repo | lenght == 0 %}present{% else %}absent{% endif %}"
￼    update_cache: yes
￼  when: (not ansible_check_mode)
  with_dict:
    5: {{ datadog_agent5_apt_repo }}
    6: {{ datadog_agent6_apt_repo }}
    7: {{ datadog_agent7_apt_repo }}
```

Same thing for Suse and Centos.

tasks/pkg-suse.yml
```diff
@@ -9,20 +9,20 @@
     args:
       warn: no
     when: not ddkey.stat.exists
-  when: ansible_distribution_version|int == 11
+  when: datadog_agent_major_version < 7 and ansible_distribution_version|int == 11
```
hush-hush: Why aren't we importing the key when using A7 ?

tasks/pkg-windows.yml
```diff
@@ -14,11 +14,11 @@
     Set-ExecutionPolicy Bypass -Scope Process -Force
     &$env:temp\fix_6_14.ps1
 
-- include_tasks: win_agent_6_latest.yml
-  when: not datadog_agent5 and (datadog_agent_version | length == 0)
+- include_tasks: win_agent_latest.yml
+  when: datadog_agent_major_version > 5 and (datadog_agent_version | length == 0)
```
hush-hush: No need to check `datadog_agent_major_version` here and in the next task since we fail the run line 1 is A5 is set.

tasks/win_agent_latest.yml
```diff
@@ -0,0 +1,11 @@
+---
+
+- name: (Custom) Set agent download filename to latest
+  set_fact:
+    dd_download_url: "{{ datadog_windows_download_url }}"
+  when: datadog_windows_download_url | length > 0
+
+- name: Set agent download filename to latest
+  set_fact:
+    dd_download_url: "https://s3.amazonaws.com/ddagent-windows-stable/datadog-agent-{{ datadog_agent_major_version }}-latest.amd64.msi"
```
hush-hush: We should use `datadog_windows_versioned_url` here

hush-hush (changes requested):

defaults/main.yml
KSerrania: We decided to do this this way in the RFC for the config management tools changes, and that's what being done in the updated chef cookbook for now (see [these lines](https://github.com/DataDog/chef-datadog/pull/646/files#diff-04c6e90faac2675aa89e2176d2eec7d8R127-R128) for example), as this is what we were doing before (except with a boolean `datadog_agent5` instead of an integer value). If you think we should add logic to avoid duplication, we can discuss it.

defaults/main.yml
KSerrania: What do you mean by uninitialised var? That `datadog_apt_repo` is an empty string, or that it is not defined at all (for example if someone sets `datadog_apt_repo: ` in their vars)?

The first case is not a problem (it's the default case where we use the default repos), and for the second case, we never checked null values, but we could start doing so if you think it's necessary.

tasks/pkg-debian.yml
KSerrania: 1. What can we do about the custom repository? Once it is in an apt source file, there's no way to remove it because of the way it's done in ansible. For example, if I do two runs:
```
vars:
  datadog_apt_repo: "deb my.custom.repo stable main"
```

```
vars:
  datadog_apt_repo: ""
```
then `deb my.custom.repo stable main` will stay in the source files, possibly having priority compared to the official repos (for example if we set `datadog_agent_major_version: 6` and the custom repo has Agent 7 packages).

tasks/pkg-suse.yml
KSerrania: This key isn't used to sign Agent 7 packages (we only use the new key ending with `E09422B3`), so no need to import the old public key.

tasks/win_agent_latest.yml
KSerrania: Should we use the same `datadog_windows_versioned_url` for both latest and versioned installs then? Or should we create two separate variables?

tasks/pkg-debian.yml
KSerrania: 2. Moreover, for the default repos, the logic fails if someone changes the value of `datadog_agentX_apt_repo`. For example, if someone does:
```
vars:
```
then
```
vars:
  datadog_agent7_apt_repo: "deb my.custom.repo stable 7"
```
then the original `deb https://apt.datadoghq.com stable 7` repo will remain in the source files, which may be a problem. Thus, should the default `datadog_agentX_apt_repo` values really be writable?

defaults/main.yml
hush-hush: I thought it was one or the other from memory of the RFC. It doesn't make sens for ansible to duplicate both and it's error prone: what if I set major to 6 and pin 7 with the version.

I'm sure it make sens for chef, but with ansible it complexify everything for no gain.

defaults/main.yml
hush-hush: I'm talking about this https://github.com/DataDog/ansible-datadog/pull/220/files?file-filters%5B%5D=.j2&file-filters%5B%5D=.yml#diff-3cd6ba6341787eb8c9742f9cd62b4c10R46 or this https://github.com/DataDog/ansible-datadog/pull/220/files?file-filters%5B%5D=.j2&file-filters%5B%5D=.yml#diff-3cd6ba6341787eb8c9742f9cd62b4c10R39 where we hardcode default in the task instead of using the defaults/main.yml

tasks/pkg-debian.yml
hush-hush: Good points. I think there is 2 cases here:
- the default one where they don't overwrite any repo URL: it should work outside the box
- the case where they change the URLs with custom ones. We can't do much about this case I think since ansible abstract the final files. People doing this will have to handle the revert themselves I think. WDYT ?

tasks/pkg-suse.yml
hush-hush: I see, could we add a comment ?

tasks/win_agent_latest.yml
hush-hush: The file names are different `datadog-agent-6-latest.amd64.msi` vs `ddagent-cli-X.Y.Z.msi` and having 2 different variables is useful (for example when we need to pin a specific version as latest). I'll say let's use 2 vars.

defaults/main.yml
albertvaka: Take the new `agent_major_version` as the previous `datadog_agent5` variable we already had, just that now it can have three values instead of two so it can't be a boolean anymore. We aren't adding more complexity, only changing a type. If you use the default (agent 7) you don't even need to specify it, like it was the case before.

I can see the duplicity in specifying the version, but it's not worse than the previous boolean we had. No approach is perfect and there are pros and cons to both, but we discussed this a while back and got to this conclusion.

For example, having a single variable doesn't let you pin "latest agent 6", which would be a regression (unless we allowed to use `6.x` or something like that, but that adds a lot more complexity). This change also keeps the code changes to the minimum, since it's only changing the boolean check to an integer check.

Also, keeping Chef, Puppet and Ansible similar is also a benefit IMO.

tasks/win_agent_latest.yml
KSerrania: To specify a specific version as latest, we already have the `datadog_windows_download_url` variable (that we could rename, it's not that explicit that it overrides the default latest), used above. Would that be enough? Otherwise we'll have 3 variables, with some overlap (one to override the full path to latest, one to override the path prefix to latest, and another one to override the path prefix to a versioned msi), which may be confusing to users.

defaults/main.yml
KSerrania: What I meant is: when would we have an uninitialised var, and when would that be a problem?

tasks/pkg-debian.yml
KSerrania: That's fine, but I fail to see the purpose of having multiple `datadog_agentX_apt_repo` variables available, since you can only target one version in a role at the same time (via the `agent_major_version` variable). The only benefit I see would be if someone repeatedly switches between major agent versions while using custom repos and doesn't want to change the `datadog_apt_repo` variable each time.

defaults/main.yml
hush-hush: I agree it's not worth than before but since we're going to bump the major we have an opportunity to make is way better ! It would not be a lot of work to deduce the major from the pinned version (literally one ansible task) and would greatly improve our user experience. Also this would be the perfect timing to abstract the difference in the version format between each OS (`6.15.0-1`, `1:6.15.0-1`, `6.15.0` ...).

And I don't think we should have Chef way of doing things in Ansible for example. We should be aiming at a native way of doing things. Users don't use chef and ansible and puppet at the same time, they only use one of those. Aligning behavior between all of those tools is a improvement for us but a drawback for our users.

In the end I think having `agent_major_version` and `agent_version` is a good idea (hence me validating the RFC), I just think we should deduce one when the other is pinned. Respecting the RFC but also making it intuitive to use.

tasks/win_agent_latest.yml
hush-hush: My main point here is to avoid having hard-coded information in the tasks. Using the defaults makes our role more flexible and easier to customize of users.

I think we only need 2:
- 1 to overwrite latest
- 1 for the versioned path

defaults/main.yml
hush-hush: So I would like to avoid having the repo hard-coded in the task. Also we had issue in the past (related to system-probe for example) where a variable initialized to an empty string in the default cause the role to fail on some edge case after a rework.

We should just move the value from the task to the default and use them. That way we don't have to check id the value is empty in the task.

tasks/pkg-debian.yml
hush-hush: advantages are:
- it allow us to have 1 task for all the repos instead of 1 per repo
- the role need to support rollback to previous major (we need to remove 7 repo when using 6 or 5 for example)

defaults/main.yml
albertvaka: What do we do if the user specifies both, then? Which one should take precedence? Isn't that more unpredictable from a user perspective?

I don't think we are doing things "the Chef way" on Ansible: this is not the Chef way either. This is a tool-agnostic solution that we decided to use across all config management tools and I have a strong opinion on not having Ansible be different from the others.

Unifying version format (`6.15.0`, `1:6.15.0-1`...) is something that we could have done but didn't make it into the proposal, so I also think it is late to propose it. The change should be backwards compatible if possible and consistent across all config management tools (that's why we made a unified proposal) and not something to add for Ansible only in a code review.

defaults/main.yml
albertvaka: Can you give a use case where the user would have to change the repo URL? I can't see why this is needed.

tasks/pkg-debian.yml
albertvaka: Valid points, but isn't there a more beautiful way to achieve this than having N variables?

defaults/main.yml
olivielpeau: ok, so I see 2 things being discussed here:

* inferring `agent_major_version` from `agent_version`: we can do this later if we think it's worth it, it would not be a breaking change anyway, so I don't think we should try to implement this now.
* abstracting out the difference in the version format between OSes: we could consider doing this indeed, it could make life easier for users. I'd find a way to make this change in a non-breaking way though, to make the upgrade path easier. Maybe something for a future minor version.

tasks/pkg-debian.yml
olivielpeau: some thoughts here:

1. we should start using the `filename` parameter of `datadog_apt_repo` (https://docs.ansible.com/ansible/latest/modules/apt_repository_module.html) and set it to something stable (a hardcoded value like `datadog_agent` should be fine), so that the name of the `.list` file is stable whether we're installing 5, 6 or 7. This will allow rollbacks. I don't think we need to have multiple variables for `datadog_apt_repo` with this solution.
2. to ease the upgrade path from the previous major version of the role, we should ensure that the `.list` file that was added by the previous major version of the role is properly absent. Otherwise once we start using `filename`, the `.list` file will have a completely different name compared to before, and users will end up with a duplicate `.list` file (which may even prevent rollbacks/introduce weird behaviors)

I think we should have a similar logic for CentOS and SUSE (haven't looked at specifics of how it works right now though)

KSerrania: About 1. 
The name of the `.list` file is already "stable", as it only depends on the url of the repo that's being added (in our case, by default the `.list` file is named `apt_datadoghq_com.list`). Moreover, the filename isn't really relevant, as when you set the repo as absent, ansible removes lines which exactly match the repo we want absent in all files in the sources folder.

Using the following implementation (one `datadog_agent_repo` var used to set a custom repo, defaults are hardcoded):
```
- name: Ensure Datadog repository is up-to-date
  apt_repository:
    repo: "{{ item.value }}"
    state: "{% if item.key == datadog_agent_major_version and datadog_apt_repo | length == 0 %}present{% else %}absent{% endif %}"
    update_cache: yes
  when: (not ansible_check_mode)
  with_dict: "{
  5: 'deb https://apt.datadoghq.com stable main',
  6: 'deb https://apt.datadoghq.com stable 6',
  7: 'deb https://apt.datadoghq.com stable 7' }"

- name: (Custom) Ensure Datadog repository is up-to-date
  apt_repository:
    repo: "{{ datadog_apt_repo }}"
    state: present
    update_cache: yes
  when: (datadog_apt_repo | length > 0) and (not ansible_check_mode)
``` 
[1]

we have the following behavior:
- if `datadog_agent_major_version == 7`: `deb https://apt.datadoghq.com stable 7` is added, `deb https://apt.datadoghq.com stable 6` and `deb https://apt.datadoghq.com stable main` are removed from all source files.
- if `datadog_agent_major_version == 6 or 5`: same behavior, simply switch the lines added / removed
- if `datadog_apt_repo` is set: `deb https://apt.datadoghq.com stable 7`, `deb https://apt.datadoghq.com stable 6` and `deb https://apt.datadoghq.com stable main` are removed from the source files. `{{ datadog_apt_repo }}` is added to the `{{ datadog_apt_repo_url }}.list` file.
=> The problematic scenario is thus the following: someone sets a custom repo, does a run, and then removes it; the custom repo remains here because there is no logic to remove it. However, this problem can be easily solved by using the `filename` parameter for the custom repo. Then, at the start of each run, we delete this file.

About 2.
This implementation also has the benefit of automatically working when upgrading: the official repositories that are in the absent state will be automatically removed from the source lists. The only repositories that may be left over after an upgrade are the custom repository lists from past runs (pre-upgrade), but we can't do anything about them anyway since we don't track them.

[1] Why don't we put the custom repo with the others in the same task?
That's because we need to check for emptiness before doing anything, as `apt_repository` will fail if given an empty string as `repo`.

----
>  I don't think we need to have multiple variables for datadog_apt_repo with this solution.

The question here is not about whether we _need_ multiple `datadog_apt_repo` variables or not: both solutions are doable (I presented a solution which uses only one, but there is also a solution using one var per Agent version). It's: do we _want_ to have hardcoded defaults and one override variable, or one variable per Agent version?

defaults/main.yml
hush-hush: Regarding backward compatibility we're going to break the behavior anyway, by installing 7 by default (and therefor bump the major). This mean a manual update of the role for user. We won't be able to break backward compatibility again later (or we would have to bump the major again). I think now is a good time to implement such simple changes.

Regarding:
> What do we do if the user specifies both, then? Which one should take precedence? Isn't 
> that more unpredictable from a user perspective?

Simple: we raise an error or ignore `major` (I vote for raising an error). They either use `major` to pin latest or use `version` to pin a specific version.

tasks/pkg-debian.yml
hush-hush: Olivier input looks great, having 1 file that contains our repo (even with custom URL) would simplify everything. We could then simplify things to:

- first task removing that file (`/etc/apt/sources.list.d/datadog_ansible.list` ?)
- second task adding a new repo to that file

Yum has a `file` option for the same purpose: (https://docs.ansible.com/ansible/latest/modules/yum_repository_module.html?highlight=yum_repository) and we're using a template for SUSE anyway.

defaults/main.yml
KSerrania: As discussed, this will be implemented in a separate PR.

README.md
```diff
@@ -36,13 +37,50 @@ Supports most Debian and RHEL-based Linux distributions, and Windows.
 ansible-galaxy install Datadog.datadog
 ```
 
+## Role upgrade from v2 to v3
+
+The `datadog_agent_major_version` variable has been introduced, to tell the module which major version of the Agent will be installed. by default, it is set to `7`.
```
hush-hush:
```suggestion
The `datadog_agent_major_version` variable has been introduced, to tell the module which major version of the Agent will be installed, `7` by default. 
```

README.md
```diff
@@ -36,13 +37,50 @@ Supports most Debian and RHEL-based Linux distributions, and Windows.
...
 ```
 
+## Role upgrade from v2 to v3
+
+The `datadog_agent_major_version` variable has been introduced, to tell the module which major version of the Agent will be installed. by default, it is set to `7`. 
+To install Agent 5, set it to `5`. To install Agent 6, set it to `6`.
+
+The `datadog_agent5` variable is thus obsolete and has been removed.
+
+The `datadog_agent_version` variable, if set, must use the major version set in `datadog_agent_major_version`.
```
hush-hush: This is no longer true

README.md
```diff
@@ -36,13 +37,50 @@ Supports most Debian and RHEL-based Linux distributions, and Windows.
...
+
+`datadog_agent_major_version` | Default apt repository | Default yum repository | Default zypper repository |
+------------------------------|------------------------|-----------------------|---------------------------|
+5                             | deb https://apt.datadoghq.com stable main | https://yum.datadoghq.com/rpm | https://yum.datadoghq.com/suse/rpm |
+6                             | deb https://apt.datadoghq.com stable 6 | https://yum.datadoghq.com/stable/6 | https://yum.datadoghq.com/suse/stable/6 |
+7                             | deb https://apt.datadoghq.com stable 7 | https://yum.datadoghq.com/stable/7 | https://yum.datadoghq.com/suse/stable/7 |
+
+To override the default behavior, set the `datadog_apt_repo`, `datadog_yum_repo`, or `datadog_zypper_repo` variables to something else than an empty string.
+
+If you were previously using the Agent 5 variables `datadog_agent5_apt_repo`, `datadog_agent5_yum_repo`, or `datadog_agent5_zypper_repo` to set custom Agent 5 repositories, use `datadog_apt_repo`, `datadog_yum_repo`, or `datadog_zypper_repo`(with `datadog_agent_major_version` set to `5`) instead.
```
hush-hush:
```suggestion
If you were previously using the Agent 5 variables `datadog_agent5_apt_repo`, `datadog_agent5_yum_repo`, or `datadog_agent5_zypper_repo` to set custom Agent 5 repositories, use `datadog_apt_repo`, `datadog_yum_repo`, or `datadog_zypper_repo`(with `datadog_agent_major_version` set to `5` or `datadog_agent_version` pinned to a specific version of agent 5) instead.
```

README.md
```diff
@@ -36,13 +37,50 @@ Supports most Debian and RHEL-based Linux distributions, and Windows.
 ansible-galaxy install Datadog.datadog
 ```
 
+## Role upgrade from v2 to v3
```
hush-hush: this section should be after the main doc `## Role Variables`.

README.md
```diff
@@ -4,7 +4,8 @@
 [![Ansible Galaxy](http://img.shields.io/badge/galaxy-Datadog.datadog-660198.svg)](https://galaxy.ansible.com/Datadog/datadog/)
 [![Build Status](https://travis-ci.org/DataDog/ansible-datadog.svg?branch=master)](https://travis-ci.org/DataDog/ansible-datadog)
 
-Install and configure Datadog Agent & checks. Starting with version `2` of this role Datadog Agent version 6 is installed by default (instead of version 5).
+Install and configure Datadog Agent & checks.
+The version `3` of this role installs the Datadog Agent version 7 by default.
```
hush-hush: Could you also update the TOC of the README line 12.

README.md
```diff
@@ -36,13 +37,50 @@ Supports most Debian and RHEL-based Linux distributions, and Windows.
...
-| `datadog_agent_version`                                                                                                                         | The pinned version of the Agent to install (optional, but highly recommended). Examples: `1:6.0.0-1` on apt-based platforms, `6.0.0-1` on yum-based platforms, `6.0.0` on Windows platforms. **Note** Downgrades are not supported on Windows platforms. |
+| `datadog_agent_major_version`                                                                                                                         | The major version of the Agent which will be installed. Its possible values are 5, 6, 7. If both this and `datadog_agent_version` are not set, then `datadog_agent_major_version` defaults to 7. If this is not set but `datadog_agent_version` is set, then `datadog_agent_major_version` is set to the major version provided in `datadog_agent_version`. |
```
hush-hush:
```suggestion
| `datadog_agent_major_version`                                                                                                                         | The major version of the Agent which will be installed. Its possible values are 5, 6 or 7 (default to 7). If set `datadog_agent_version` will take precedence, otherwise the latest version of the specified major will be installed. Setting `datadog_agent_major_version` is not needed if `datadog_agent_version` is used. |
```

README.md
```diff
@@ -36,13 +37,50 @@ Supports most Debian and RHEL-based Linux distributions, and Windows.
...
+| `datadog_agent_version`                                                                                                                         | The pinned version of the Agent to install (optional, but highly recommended). Examples: `1:6.0.0-1` on apt-based platforms, `6.0.0-1` on yum-based platforms, `6.0.0` on Windows platforms. If both this variable and `datadog_agent_major_version` are set, they must be compatible (the major version in `datadog_agent_version` has to be `datadog_agent_major_version`). **Note** Downgrades are not supported on Windows platforms. |
```
hush-hush:
```suggestion
| `datadog_agent_version`                                                                                                                         | The pinned version of the Agent to install (optional, but highly recommended). Example: `7.15.0`. Using `datadog_agent_major_version` is not needed is `datadog_agent_version` is used.. **Note** Downgrades are not supported on Windows platforms. |
```

I think we should first document `datadog_agent_version` then `datadog_agent_major_version` as we advice people to use `datadog_agent_version` over `datadog_agent_major_version`.

I don't think we need to document that both `datadog_agent_version` and `datadog_agent_major_version` should be compatible. The main idea is not to use both at the same time.

README.md
```diff
@@ -90,21 +128,24 @@ and the role will install `1:7.16.0-1` on Debian- and SUSE-based systems, `7.16.
 
 ## Agent 5 (older version)
 
-This role includes support for Datadog Agent version 5 for linux only. To install Agent 5, you need to:
+This role includes support for Datadog Agent version 5 for Linux only. To install Agent 5, you need to:
 
-1. Set `datadog_agent5` parmeter to `true`.
+1. Set `datadog_agent_major_version` parmeter to `5`.
```
hush-hush: We can simplify now:

```
Either set `datadog_agent_major_version` to 5 to install the latest version of Agent 5 or set `datadog_agent_version` to an existing Agent v5.
```

README.md
```diff
@@ -90,21 +128,24 @@ and the role will install `1:7.16.0-1` on Debian- and SUSE-based systems, `7.16.
...
 2. Set `datadog_agent_version` to an existing Agent v5 version or leave it empty to always install the latest version (`5.*`).
 
-To downgrade from Agent v6 to Agent v5, you need to (**on centos this will only work with ansible 2.4 and up**):
+## Agent major version downgrades
 
-1. Set `datadog_agent5` to `true`.
+To downgrade from Agent v6 or v7 to Agent v5, you need to (**on centos this will only work with ansible 2.4 and up**):
```
hush-hush: We should also add a work about downgrading not being supported on Windows.

README.md
```diff
@@ -90,21 +128,24 @@ and the role will install `1:7.16.0-1` on Debian- and SUSE-based systems, `7.16.
...
+
+1. Set `datadog_agent_major_version` to `5`.
 2. Pin `datadog_agent_version` to an existing Agent 5 version.
 3. Set `datadog_agent_allow_downgrade` to `yes`.
 
-Variables:
+Likewise, to downgrade from Agent v7 to Agent v6, you need to (**on centos this will only work with ansible 2.4 and up**):
```
hush-hush: Same here for Windows.

README.md
```diff
@@ -90,21 +128,24 @@ and the role will install `1:7.16.0-1` on Debian- and SUSE-based systems, `7.16.
...
 
-This role includes support for Datadog Agent version 5 for linux only. To install Agent 5, you need to:
+This role includes support for Datadog Agent version 5 for Linux only. To install Agent 5, you need to:
 
-1. Set `datadog_agent5` parmeter to `true`.
+1. Set `datadog_agent_major_version` parmeter to `5`.
 2. Set `datadog_agent_version` to an existing Agent v5 version or leave it empty to always install the latest version (`5.*`).
 
-To downgrade from Agent v6 to Agent v5, you need to (**on centos this will only work with ansible 2.4 and up**):
+## Agent major version downgrades
```
hush-hush: We could merge the both paragraph since the role is much simpler now.

Something like:
```
1 .Either set `datadog_agent_major_version` to the major version you want to install the latest of that version or set `datadog_agent_version` to a specific version to (ex: 5.32.5)
2. Set `datadog_agent_allow_downgrade` to `yes`.
```

WDYT ?

README.md
```diff
@@ -225,6 +266,7 @@ Sending data to Datadog US (default) and configuring a few checks.
     - { role: Datadog.datadog, become: yes }
   vars:
     datadog_api_key: "123456"
+    datadog_agent_major_version: 6
```
hush-hush: The main Playbook example should install the latest version. Let's add another example to showcase `datadog_agent_major_version` and maybe `datadog_agent_version`.

tasks/main.yml
```diff
@@ -1,6 +1,6 @@
 ---
-- include_tasks: parse-version.yml
-  when: datadog_agent_version | length > 0
+
+- include_tasks: set-parse-version.yml
   run_once: true
```
hush-hush: What if I set different version per host ? I think using `run_once` here would introduce a buggy behavior, no ?

tasks/pkg-debian.yml
```diff
@@ -31,21 +36,29 @@
...
     - "deb http://apt.datadoghq.com/ stable main"
     - "deb http://apt.datadoghq.com/ stable 6"
+    - "deb http://apt.datadoghq.com/ stable 7"
   when: not ansible_check_mode
 
 - name: Ensure Datadog repository is up-to-date
   apt_repository:
-    repo: "{{ datadog_apt_repo }}"
-    state: "{% if datadog_agent5 %}absent{% else %}present{% endif %}"
+    filename: datadog_agent
```
hush-hush: I think we should prefix/suffix the filename by `ansible` to avoid clash with any user created file. WDYT ?

This would also allow us to detect who dropped that repo file when troubleshooting user install.

tasks/pkg-debian.yml
```diff
@@ -31,21 +36,29 @@
...
+    repo: "{{ item.value }}"
+    state: "{% if item.key == datadog_agent_major_version|int and datadog_apt_repo | length == 0 %}present{% else %}absent{% endif %}"
     update_cache: yes
-  when: not ansible_check_mode
+  when: (not ansible_check_mode)
+  with_dict: "{
```
hush-hush: quote are not needed, we can directly use a yaml dict:
```
  with_dict:
    5: {{ datadog_agent5_apt_repo }}
    6: {{ datadog_agent6_apt_repo }}
    7: {{ datadog_agent7_apt_repo }}
```

tasks/pkg-redhat.yml
```diff
@@ -11,25 +11,32 @@
...
-    baseurl: "{{ datadog_yum_repo }}"
+    baseurl: "{{ item.value }}"
     enabled: yes
     gpgcheck: yes
-    gpgkey: "{{ datadog_yum_gpgkey }}"
-    state: "{% if datadog_agent5 %}absent{% else %}present{% endif %}"
+    gpgkey: "{% if datadog_agent_major_version|int == 7 %}{{ datadog_yum_gpgkey_e09422b3 }}{% else %}{{ datadog_yum_gpgkey }}{% endif %}"
+    state: "{% if item.key == datadog_agent_major_version|int and datadog_yum_repo | length == 0 %}present{% else %}absent{% endif %}"
+  when: (not ansible_check_mode)
+  with_dict: "{
```
hush-hush: Same here about yaml dict

tasks/pkg-redhat.yml
```diff
@@ -11,25 +11,32 @@
     state: present
   when: not ansible_check_mode
 
-- name: Install DataDog yum repo
+- name: Install Datadog yum repo
   yum_repository:
-    name: datadog
+    name: "datadog_{{ item.key }}"
```
hush-hush: same here, `datadog_ansible_{{ item.key }}` ?

hush-hush (changes requested):

README.md
KSerrania: 👍, moreover since `datadog_agent_major_version` is not even needed if `datadog_agent_version` is set, we can remove the bit about `datadog_agent_major_version`:
 
```
1. Set `datadog_agent_version` to a specific version to downgrade to (ex: 5.32.5)
2. Set `datadog_agent_allow_downgrade` to `yes`.
```

Setting only `datadog_agent_allow_downgrade` and `datadog_agent_major_version` does not work, you need to specify a minor version for the downgrade to work.

README.md
```diff
@@ -225,7 +263,7 @@ Sending data to Datadog US (default) and configuring a few checks.
     - { role: Datadog.datadog, become: yes }
   vars:
     datadog_api_key: "123456"
-    datadog_agent_version: "1:6.13.0-1" # for apt-based platforms, use a `6.13.0-1` format on yum-based platforms and `6.13.0` for Windows
+    datadog_agent_version: "1:7.16.0-1" # apt-based format; the role will deduce and use the `7.16.0-1` format on yum-based platforms and `7.16.0` on Windows
```
hush-hush: We should just use the `7.16.0` format. The rest is only here to be backward compatible.

hush-hush (approved):
I added a small comment. Feel free to merge after addressing it !

README.md
```diff
@@ -4,13 +4,17 @@
 [![Ansible Galaxy](http://img.shields.io/badge/galaxy-Datadog.datadog-660198.svg)](https://galaxy.ansible.com/Datadog/datadog/)
 [![Build Status](https://travis-ci.org/DataDog/ansible-datadog.svg?branch=master)](https://travis-ci.org/DataDog/ansible-datadog)
 
-Install and configure Datadog Agent & checks. Starting with version `2` of this role Datadog Agent version 6 is installed by default (instead of version 5).
+Install and configure Datadog Agent & checks.
+The version `3` of this role installs the Datadog Agent version 7 by default.
```
hush-hush: Just realize that it's version 4 ! the role is already in version 3.3.0

hush-hush (changes requested):

README.md
KSerrania: I just realized too! I'm going to change v3 -> v4 and v2 -> v3.

rromanchuk:
@KSerrania i'm going to give your fork a go, thanks for this

tasks/pkg-debian.yml
```diff
@@ -31,21 +36,28 @@
...
+    - "deb http://apt.datadoghq.com/ stable 7"
   when: not ansible_check_mode
 
 - name: Ensure Datadog repository is up-to-date
   apt_repository:
-    repo: "{{ datadog_apt_repo }}"
-    state: "{% if datadog_agent5 %}absent{% else %}present{% endif %}"
+    filename: ansible_datadog_agent
+    repo: "{{ item.value }}"
+    state: "{% if item.key == datadog_agent_major_version|int and datadog_apt_repo | length == 0 %}present{% else %}absent{% endif %}"
```
olivielpeau, line 46: does this actually work if Agent v5 or v6 is selected? If I understand the logic correctly, using `with_dict` will create 3 tasks (one for each major version), which will in turn either delete or create the `/etc/apt/sources.list.d/ansible_datadog_agent.list` file. So I'd expect the last defined task to always "win", i.e. always the v7 task since it's the last item of `with_dict`.

Instead of having this logic in `state`, shouldn't the logic be in the `when` condition, and `state` always be set to `present`? I believe `when` is evaluated for each item of `with_dict`, but I may be wrong or missing something.

tasks/pkg-debian.yml
```diff
@@ -23,6 +23,11 @@
     state: present
   when: datadog_apt_key_url_new is defined
 
+- name: Remove previous datadog apt list file
+  file:
+    path: /etc/apt/sources.list.d/ansible_datadog_agent.list
+    state: absent
```
olivielpeau, line 29: not sure I understand what this task is for: if I understand correctly the tasks below will update the `/etc/apt/sources.list.d/ansible_datadog_agent.list` anyway, so why do we need to delete it here?

README.md
```diff
@@ -44,28 +48,65 @@ ansible-galaxy install Datadog.datadog
...
-| `datadog_checks`                                                                                                                                | YAML configuration for agent checks to drop into: <br> - `/etc/datadog-agent/conf.d/<check_name>.d/conf.yaml` for Agent v6. <br> - `/etc/dd-agent/conf.d` for Agent v5.                                                                                  |
-| `datadog_config`                                                                                                                                | Settings to place in the main Agent configuration file: <br> - `/etc/datadog-agent/datadog.yaml` for Agent v6 <br> - `/etc/dd-agent/datadog.conf` for Agent v5 (under the `[Main]` section).                                                               |
+| `datadog_agent_version`                                                                                                                         | The pinned version of the Agent to install (optional, but highly recommended). Example: `7.15.0`. Setting `datadog_agent_major_version` is not needed if `datadog_agent_version` is used. **Note** Downgrades are not supported on Windows platforms. |
```
olivielpeau: `7.15.0` doesn't exist:

```suggestion
| `datadog_agent_version`                                                                                                                         | The pinned version of the Agent to install (optional, but highly recommended). Example: `7.16.0`. Setting `datadog_agent_major_version` is not needed if `datadog_agent_version` is used. **Note** Downgrades are not supported on Windows platforms. |
```

olivielpeau (commented):
Left a question inline, and a few comments, LGTM overall.

One more suggestion: In `defaults/main.yml`, I think we should more clearly separate the internal variables that are not part of the stable API of the role, from the variables that are part of the stable API. Typically, put all the internal variables at the end of the file in a section clearly separated with a very visible comment.

tasks/pkg-debian.yml
KSerrania: As mentioned in the docs: https://docs.ansible.com/ansible/latest/modules/apt_repository_module.html, the state only applies to the source string, not the file: if `absent`, ansible will make sure that the `repo` string is not present in any source file, the file itself won't be erased.

tasks/pkg-debian.yml
KSerrania: The custom repository won't get deleted if this is not present. Moreover, if the internal variables (`datadog_agentX_yyyyy_repo`) are modified, the previous values won't get removed from the source files, which may cause issues. IMO it's safer to explicitly remove the list file.

It does have the downside of making the role not idempotent though, so this is up for discussion.

tasks/pkg-debian.yml
olivielpeau: ok, thanks for the explanation 👍

tasks/pkg-debian.yml
olivielpeau: Ok, thanks for the explanation, makes sense. Indeed, the idempotency would be nice to have, but not blocking for this PR.

In a future version, we may want to revisit using the `apt_repository` task. For our use case, I think all of this logic would be easier to implement with one single `file` task that would trigger `apt-get update` on the datadog repo whenever needed. Not necessary for now though.

olivielpeau (approved):

PR #221, opened 2019-11-12
Blacklist version 6.14.0 and 6.14.1 version on Windows

hush-hush:
The Agent versions you pinned (6.14.0 or 6.14.1) have been blacklisted
on Windows. See http://dtdg.co/win-614-fix.  Ansible now refuse to
install those version, default to 6.14.2 on Windows and will run the fix
PowerShell script in case 6.14.0/1 was installed.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -57,8 +57,11 @@ datadog_skip_running_check: false
 datadog_agent_allow_downgrade: no
 
 # default datadog windows download url
-datadog_windows_download_url: "https://s3.amazonaws.com/ddagent-windows-stable/datadog-agent-6-latest.amd64.msi"
+# we're pinning 6.14.2 instead of latest for now. See http://dtdg.co/win-614-fix for more details.
+#datadog_windows_download_url: "https://s3.amazonaws.com/ddagent-windows-stable/datadog-agent-6-latest.amd64.msi"
+datadog_windows_download_url: "https://s3.amazonaws.com/ddagent-windows-stable/ddagent-cli-6.14.2.msi"
 datadog_windows_versioned_url: "https://s3.amazonaws.com/ddagent-windows-stable/ddagent-cli"
+datadog_windows_614_fix_script_url: "https://s3.amazonaws.com/ddagent-windows-stable/scripts/fix_6_14.ps1"
 
 # Override to change the name of the windows user to create
 datadog_windows_ddagentuser_name: ""
--- a/tasks/pkg-windows.yml
+++ b/tasks/pkg-windows.yml
@@ -4,6 +4,16 @@
     msg: "The Datadog ansible role does not currently support Agent 5"
   when: datadog_agent5
 
+- name: Download windows datadog agent 614 fix script
+  win_get_url:
+    url: "{{ datadog_windows_614_fix_script_url }}"
+    dest: '%TEMP%\fix_6_14.ps1'
+
+- name: Run 6.14.0/1 PowerShell fix
+  win_shell: |
+    Set-ExecutionPolicy Bypass -Scope Process -Force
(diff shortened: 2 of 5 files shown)


tasks/win_agent_version.yml
```diff
@@ -1,4 +1,10 @@
 ---
+
+- name: fail the play if the previous command did not succeed
```
albertvaka: `play`?

albertvaka (approved):

tasks/win_agent_version.yml
hush-hush: Copy/paste error.

defaults/main.yml
```diff
@@ -57,7 +57,9 @@ datadog_skip_running_check: false
 datadog_agent_allow_downgrade: no
 
 # default datadog windows download url
-datadog_windows_download_url: "https://s3.amazonaws.com/ddagent-windows-stable/datadog-agent-6-latest.amd64.msi"
+# we're pinning 6.14.2 instead of latest for now. See http://dtdg.co/win-614-fix for more details.
+#datadog_windows_download_url: "https://s3.amazonaws.com/ddagent-windows-stable/datadog-agent-6-latest.amd64.msi"
```
derekwbrown, line 61: remove the commented out line?

tasks/pkg-windows.yml
```diff
@@ -4,6 +4,12 @@
     msg: "The Datadog ansible role does not currently support Agent 5"
   when: datadog_agent5
 
+- name: Run 6.14.0/1 PowerShell fix
+  win_shell: |
+    Set-ExecutionPolicy Bypass -Scope Process -Force
```
derekwbrown, line 14: does this need the (conditional) TLS settings)?

derekwbrown (approved):
Minor questions inline.  Overall looks fine.

defaults/main.yml
albertvaka: Keeping it makes it easier to revert to latest.

tasks/pkg-windows.yml
hush-hush: Where can I find the TLS setting ? I use the line from the fix page.

olivielpeau: if there's an Ansible task that allows downloading a file, I think we should prefer downloading the fix script first with that ansible task, and then use `win_shell` to run it. That way we'll be able to clearly separate errors that happen when downloading the file from errors that happen when running the script

tasks/pkg-windows.yml
```diff
@@ -4,6 +4,12 @@
...
+    ((New-Object System.Net.WebClient).DownloadFile('https://s3.amazonaws.com/ddagent-windows-stable/scripts/fix_6_14.ps1', $env:temp + '\fix_6_14.ps1'))
```
olivielpeau: I think we should allow customizing the URL we download the script from, similarly to how we allow customizing the download URL of the Agent MSIs, so that customers that use a mirror of our S3 bucket can also pull the script from their mirror

tasks/pkg-windows.yml
```diff
@@ -4,6 +4,16 @@
     msg: "The Datadog ansible role does not currently support Agent 5"
   when: datadog_agent5
 
+- name: Download windows datadog agent
```
olivielpeau: not the right name here, probably bad copy-pasta :)

hush-hush: :+1:

olivielpeau (approved):
LGTM, I'm planning to merge this PR and release v`3.2.1` on Monday

olivielpeau:
This will be shipped in `3.3.0` since we're shipping features and not only bugfixes. Merging

PR #233, opened 2019-12-05
[windows] Add README instructions for Windows hosts

KSerrania:
### What does this PR do?

Add instructions on how to make the role work with Windows hosts (`become: yes` does not work by itself on Windows).

### Motivation

Windows support.

### Additional Notes

For now, I put the section next to the playbook examples. Is there a better place?

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,7 +313,7 @@ ansible_become=no
 
 ### Using the playbook file
 
-Alternatively, if your playbook **only runs on Windows hosts**, you can do the following in the playbook:
+Alternatively, if your playbook **only runs on Windows hosts**, you can do the following in the playbook file:
 
 ```yml
 - hosts: servers
--- a/README.md
+++ b/README.md
@@ -281,25 +281,23 @@ Example for sending data to EU site:
 
 ### Making the playbook work on Windows
 
-On Windows, the `become: yes` option only works if:
-- `become_method` is set to `runas`,
-- `become_user` is set to a user with Administrator powers (for instance, `System`).
+On Windows, the `become: yes` option is not needed (and will make the role fail, as ansible won't be able to use it).
 
 Below are two methods to make the above playbook work with Windows hosts:
 
 ### Using the inventory file (recommended)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -279,6 +279,55 @@ Example for sending data to EU site:
...
+
+### Using the inventory file (recommended)
+
+The recommended way to set these options is to set then in the inventory file, by setting the `ansible_become_method` and `ansible_become_user` options:
+
+```ini
+[servers]
+linux1 ansible_host=127.0.0.1
+linux2 ansible_host=127.0.0.2
+windows1 ansible_host=127.0.0.3 ansible_become_method=runas ansible_become_user=System
```
hush-hush: `ansible_become_user` is optional I think. By default it will use the `Winrm` user, no ?

README.md
```diff
@@ -279,6 +279,55 @@ Example for sending data to EU site:
...
+```
+
+### Using the playbook file
+
+Alternatively, if your playbook **only runs on Windows hosts**, you can set these variables directly in the playbook:
+
+```yml
+- hosts: servers
+  roles:
+    - { role: Datadog.datadog, become: yes, become_method: runas, become_user: System }
```
hush-hush: Same here I think, setting only `{ role: Datadog.datadog}` and letting ansible use the right default works. I don't think we should advice people to use System by default and let them use the default value in there version of ansible.

hush-hush (commented):
Added a few comment. WDYT ?

README.md
KSerrania: When setting `ansible_become_method` to `runas`, you have to set the `ansible_become_user`. Otherwise you get this error:
`fatal: [windows1]: FAILED! => {"msg": "No setting was provided for required configuration plugin_type: become plugin: runas setting: become_user "}`.

But actually, there's a simpler way to make the role work, without relying on a user: by setting `ansible_become=no` for Windows hosts.

hush-hush (approved):

PR #234, opened 2019-12-05
Reset windows version pinned

hush-hush:
Latest has been moved to latest and 6.14.[01] removed

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ Supports most Debian and RHEL-based Linux distributions, and Windows.
   - [Agent 5](#agent-5)
 - [Additional tasks](#additional-tasks)
 - [Known Issues and Workarounds](#known-issues-and-workarounds)
+  - [dirmngr](#dirmngr)
+  - [Datadog Agent 6.14 for Windows](#datadog-agent-614-for-windows)
 - [License](#license)
 - [Author Information](#author-information)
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -372,13 +372,12 @@ Due to a critical bug in Agent versions `6.14.0` and `6.14.1` on Windows, these
...
 **PLEASE NOTE:** ansible will fail on Windows if `datadog_agent_version` is set
-to `6.14.0` or `6.14.1`. Please use `6.14.2` instead. If the Agent version is not
-pinned, `6.14.2` will be installed by default.
+to `6.14.0` or `6.14.1`. Please use `6.14.2` or above instead.
 
 If you are updating from **6.14.0 or 6.14.1 on Windows**, we **strongly** recommend following these steps:
 
 1. Upgrade the present `Datadog.datadog` ansible role to the latest version (`>=3.3.0`)
-2. Set the `datadog_agent_version` to `6.14.2`
+2. Set the `datadog_agent_version` to `6.14.2` or above (by default the role install latest).
```
KSerrania, line 451: Is using the default role behavior also an option? In that case I'd put something like this to stress the fact that using latest is also fine:
```suggestion
2. Set the `datadog_agent_version` to `6.14.2` or above (or use the default role behavior, which will install the latest Agent 6).
```

KSerrania (approved):
LGTM, left a question about the upgrade instructions

README.md
hush-hush: If not specify the role will install the latest version. I think we should avoid mentioning `6` since `7` is around the corner.

KSerrania, line 451:
```suggestion
2. Set the `datadog_agent_version` to `6.14.2` or above (by default the role installs latest).
```

KSerrania (approved):

KSerrania:
Let's merge this and release before release the next major version

PR #239, opened 2019-12-13
Infer major version from datadog_agent_version

KSerrania:
### What does this PR do?

If `datadog_agent_version` is set, extracts the major version from it and sets `datadog_agent_major_version` with this value.
If `datadog_agent_major_version` is already set, checks that the two variables are compatible, otherwise raises an error.

### Additional notes

Due to this, we need to cast `datadog_agent_major_version` to a string (as the extracted major version is a string) in all cases.


README.md
```diff
@@ -79,8 +79,8 @@ To override the default behavior, set the `datadog_windows_download_url` variabl
...
-| `datadog_agent_version`                                                                                                                         | The pinned version of the Agent to install (optional, but highly recommended). Examples: `1:6.0.0-1` on apt-based platforms, `6.0.0-1` on yum-based platforms, `6.0.0` on Windows platforms. **Note** It must use the major version set in `datadog_agent_major_version`. **Note** Downgrades are not supported on Windows platforms. |
+| `datadog_agent_major_version`                                                                                                                         | The major version of the Agent which will be installed. Its possible values are 5, 6, 7. If both this and `datadog_agent_version` are not set, then `datadog_agent_major_version` defaults to 7. If this is not set but `datadog_agent_version` are is set, then `datadog_agent_major_version` is set to the major version provided in `datadog_agent_version`. |
```
albertvaka:
```suggestion
| `datadog_agent_major_version`                                                                                                                         | The major version of the Agent which will be installed. Its possible values are 5, 6, 7. If both this and `datadog_agent_version` are not set, then `datadog_agent_major_version` defaults to 7. If this is not set but `datadog_agent_version` is set, then `datadog_agent_major_version` is set to the major version provided in `datadog_agent_version`. |
```

albertvaka (approved):

PR #240, opened 2019-12-13
System-probe indent filter fix

### LINKED ISSUE #229: TASK [Datadog.datadog : Create system-probe configuration file] Failure
Hello, 

I started getting the following message, we didn't change anything in our config or playbook. Would that be related to any of your recent changes?

googlecompute: fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "AnsibleError: Unexpected templating type error occurred on (# Managed by Ansible\n\n{% if system_probe_config is defined and system_probe_config|length > 0 -%}\nsystem_probe_config:\n{% filter indent(width=2, first=True) %}\n{{ system_probe_config | to_nice_yaml }}\n{% endfilter %}\n{% endif %}\n): do_indent() got an unexpected keyword argument 'first'"}

Thanks

KSerrania:
### What does this PR do?

Replaces the `first` argument of the `indent` filter with the older `indentfirst`.
Fixes #229.

### Motivation

The `first` option is only supported by jinja 2.10+ which is not present on older systems (CentOS 7, Debian 8, etc.).
Using the equivalent option `indentfirst` (see [here](https://github.com/pallets/jinja/blob/4fe03dafcdf251739e1fa27843d7167a7b2f2031/jinja2/filters.py#L611-L630)) will work with all currently existing jinja 2 versions (as of this comment, up to 2.10).

### DIFF
--- a/templates/system-probe.yaml.j2
+++ b/templates/system-probe.yaml.j2
@@ -2,7 +2,14 @@
 
 {% if system_probe_config is defined and system_probe_config|length > 0 -%}
 system_probe_config:
-{% filter indent(width=2, first=True) %}
+{# The "first" option is only supported by jinja 2.10+
+  which is not present on older systems (CentOS 7, Debian 8, etc.)
+  Using the equivalent option "indentfirst" will work with
+  all currently existing jinja 2 versions (as of this comment, up to 2.10).
+  The only downside is that this will print a deprecation warning if used
+  with jinja 2.10+.
+ #}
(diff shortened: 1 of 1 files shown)


albertvaka (approved):

bsolomon1124:
@KSerrania a heads up that `indentfirst` is removed in Jinja 3.

https://github.com/pallets/jinja/blob/2.11.x/src/jinja2/filters.py#L633

PR #246, opened 2019-12-24
Update instructions to use datadog.datadog as the role name

KSerrania:
### What does this PR do?

Use the same typography as Ansible Galaxy. 

### Motivation

This avoids errors when people install using the Ansible Galaxy page instructions (thus installing `datadog.datadog`) but then follow the README (which uses `Datadog.datadog`), causing errors as the role name is not the same.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Supports most Debian and RHEL-based Linux distributions, and Windows.
 ## Installation
 
 ```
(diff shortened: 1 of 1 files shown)


albertvaka (approved):
🐱Purrfect

ghost (approved):
Great

PR #247, opened 2020-01-03
Fail with explicit message if OS is not supported by the role

KSerrania:
### What does this PR do?

Adds a task at the start of the run checking if the OS is supported by the role.
Adds a message in case it's not supported so that users can contact support to open a feature request.


kbogtob:
Looks good to me :)

PR #248, opened 2020-01-10
Change default datadog_agent group

kevinconaway:
The default group should be dd-agent, not root

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -12,7 +12,7 @@ datadog_checks: {}
 
 # default user/group
 datadog_user: dd-agent
-datadog_group: root
+datadog_group: dd-agent
 
 # agent integration variables
 integration_command_user_linux: "dd-agent"


shang-wang (approved):

KSerrania (approved):

tasks/agent-linux.yml
```diff
@@ -66,7 +66,7 @@
     src: system-probe.yaml.j2
     dest: /etc/datadog-agent/system-probe.yaml
     mode: 0640
-    owner: "root"
+    owner: "{{ datadog_user }}"
     group: "{{ datadog_group }}"
```
leeavital: I think it should be the group that changes, not the owner.  The default value for `datadog_group` in ansible is root. 


In chef the owner is root, and the group is datadog-agent.

https://github.com/DataDog/chef-datadog/blob/a3adb45dee5d821a4af860c8d20ced13ad9cc62b/recipes/system-probe.rb#L44-L45

kevinconaway: Do we need to change the group globally in the playbook or just for the system probe?

@KSerrania do you know why `datadog_group` defaults to `root`?

KSerrania: Not really, the user and group were set 4 years ago.  I found this commit: https://github.com/DataDog/ansible-datadog/commit/70b3d5ddc1cfb91f4f5a376f95582b4884e540fb which explains that the group must be root, otherwise the debian package install will fail. However, since this was 4 years ago, we should check if that's still the case. If it is not, aligning the role with our other installation methods may be a good idea.

albertvaka: We changed the group on Puppet [1] from `root` to `dd-agent` and so far we have seen no complaints. I think it should be safe to do it everywhere.

[1] https://github.com/DataDog/puppet-datadog-agent/blame/master/manifests/params.pp#L47

albertvaka (changes requested):
I think we should change the group to not be `root` instead of the owner.

olivielpeau:
FWIW, these config mgmt tools used to set the `root` group because that's what the agent package itself did a _long time_ ago, before v5.12: see https://github.com/DataDog/dd-agent-omnibus/pull/101 if you like archeology. So it's perfectly fine to change the group to `dd-agent` now.

Note that it's important for the `system-probe.yaml` config file to **not** be writeable by the `dd-agent` user, since it's read by a `root` process. I think we should change the file's mode to `0440`, similar to what the agent package itself does here: https://github.com/DataDog/datadog-agent/blob/7785d5b66e3f4122f4938b876bd66b548c2ab8c9/omnibus/package-scripts/agent/postinst#L65-L69

kevinconaway:
>  I think we should change the file's mode to 0440, similar to what the agent package itself does here

In both the chef recipe and here in ansible, we have it as `0640` which should suffice since it is owned by root.

leeavital (approved):

olivielpeau:
> In both the chef recipe and here in ansible, we have it as 0640 which should suffice since it is owned by root.

OK, just a note then: for consistency it'd be nice to change the `postinst` script of the package so that it applies the same ownership and permissions to the system-probe's config files as our config mgmt tools.

PR #249, opened 2020-01-16
Ensure that system-probe is stopped if it is disabled or not installed

kevinconaway:
If the system probe is not installed or is explicitly disabled, we should ensure that the service is stopped. 

This covers the use case where the probe had been previously enabled and then disabled.

### DIFF
--- a/tasks/agent-linux.yml
+++ b/tasks/agent-linux.yml
@@ -89,6 +89,13 @@
     enabled: yes
   when: not datadog_skip_running_check and datadog_enabled and not ansible_check_mode and datadog_sysprobe_enabled
 
+- name: Ensure datadog-agent-sysprobe is stopped if disabled or not installed
+  service:
+    name: datadog-agent-sysprobe
+    state: stopped
+    enabled: no
+  when: not datadog_skip_running_check and (not datadog_enabled or not datadog_sysprobe_enabled)
+
 - name: Ensure datadog-agent, datadog-agent-process and datadog-agent-trace are not running
   service:
     name: "{{ item }}"


shang-wang (approved):

tasks/agent-linux.yml
```diff
@@ -89,6 +89,13 @@
     enabled: yes
   when: not datadog_skip_running_check and datadog_enabled and not ansible_check_mode and datadog_sysprobe_enabled
 
+- name: Ensure datadog-agent-sysprobe is stopped if disabled or not installed
+  service:
+    name: datadog-agent-sysprobe
+    state: stopped
+    enabled: no
+  when: not datadog_skip_running_check and (not datadog_enabled or not datadog_sysprobe_enabled)
```
KSerrania, line 97: Let's prevent this from running when ansible is in check mode (same as the running check above):

```suggestion
  when: not datadog_skip_running_check and not ansible_check_mode and (not datadog_enabled or not datadog_sysprobe_enabled)
```

kevinconaway: The reason I omitted that is the task to [stop the main agent](https://github.com/DataDog/ansible-datadog/blob/27b302ea295fea2814c5b6122419c5f7fe9ddb2c/tasks/agent-linux.yml#L104) does not include it.

Does it still make sense to include?

KSerrania: I'd say this is an error, as this means the role may do something when run in check mode (for instance, if an Agent was running on the target host, the role may stop it).
But for now, let's keep the role behaviour coherent and leave this out. I'll investigate this in another PR.

KSerrania (approved):

PR #250, opened 2020-01-17
pkg-debian: fix repo idempotence

jharley:
While doing development with Molecule on 18.04, I discovered that `pkg-debian.yml` was written in such a way that runs could never be idempotent. This changes the repository file name logic to basically match what is happening in `pkg-redhat.yml` and makes the role (more?) idempotent.

### DIFF
--- a/tasks/pkg-debian.yml
+++ b/tasks/pkg-debian.yml
@@ -41,7 +41,7 @@
 
 - name: Ensure Datadog repository is up-to-date
   apt_repository:
-    filename: ansible_datadog_agent
+    filename: "ansible_datadog_{{ item.key }}"
     repo: "{{ item.value }}"
     state: "{% if item.key == datadog_agent_major_version|int and datadog_apt_repo | length == 0 %}present{% else %}absent{% endif %}"
     update_cache: yes
@@ -53,7 +53,7 @@
 
 - name: (Custom) Ensure Datadog repository is up-to-date
   apt_repository:
-    filename: ansible_datadog_agent
+    filename: ansible_datadog_custom
     repo: "{{ datadog_apt_repo }}"
     state: present
     update_cache: yes


tasks/pkg-debian.yml
```diff
@@ -53,7 +53,7 @@
 
 - name: (Custom) Ensure Datadog repository is up-to-date
   apt_repository:
-    filename: ansible_datadog_agent
+    filename: ansible_datadog_custom
```
KSerrania, line 56: We tried doing that before, but this method has a problem in the following situation:
- Let's say you do a first run with the following configuration:
```
datadog_apt_repo: "deb https://my.custom.repo stable main"
```
Then, after this run, you have:
```
$ find /etc/apt/sources.list.d/ -type f -print -exec cat {} \;
/etc/apt/sources.list.d/ansible_datadog_custom.list
deb https://my.custom.repo stable main
```

- Then, on the following run, you decide to use an official repo and use the following configuration:
```
```
Because of this, the`(Custom)` task will be skipped, so the custom repository won't be added, but _it won't be removed either_, and you'll end up with:
```
$ find /etc/apt/sources.list.d/ -type f -print -exec cat {} \;
/etc/apt/sources.list.d/ansible_datadog_7.list
deb https://apt.datadoghq.com/ stable 7
/etc/apt/sources.list.d/ansible_datadog_custom.list
deb https://my.custom.repo stable main
```
and in this situation, the role may have an unexpected behavior as you can't know in advance from which source the installed package will come from.

----------------

Another problematic situation is the following:
- First run with `datadog_apt_repo: "deb https://my.custom.repo stable main"`
- Second run with `datadog_apt_repo: "deb https://my.custom.repo2 stable main"`
Then you end up with:
```
$ find /etc/apt/sources.list.d/ -type f -print -exec cat {} \;
/etc/apt/sources.list.d/ansible_datadog_custom.list
deb https://my.custom.repo stable main
deb https://my.custom.repo2 stable main
```

In this case, you don't know from which custom repository will the package come from.

----------------

I think the following fix would work for both scenarios while keeping the role idempotency:
- keep your current changes
- remove lines 26-29 (they're useless now)
- add the following tasks before this one:
```yaml
- name: (Custom) Initialize custom repo file deletion flag to False
  set_fact:
    datadog_remove_custom_repo_file: "False"

- name: (Custom) Check if custom repository file exists
  stat:
    path: /etc/apt/sources.list.d/ansible_datadog_custom.list
  register: datadog_custom_repo_file

- name: (Custom) Flag custom repository file for deletion if different from current repository config
  set_fact:
    datadog_remove_custom_repo_file: "{{ datadog_repo_file_contents != datadog_apt_repo }}"
  vars:
    datadog_repo_file_contents: "{{ lookup('file', '/etc/apt/sources.list.d/ansible_datadog_custom.list') }}"
  when: datadog_custom_repo_file.stat.exists

- name: (Custom) Remove Datadog custom repository file when not set or updated
  file:
    path: /etc/apt/sources.list.d/ansible_datadog_custom.list
    state: absent
  vars:
    repo_file_contents: "{{ lookup('file', '/etc/apt/sources.list.d/ansible_datadog_custom.list') }}"
  when: (not (datadog_apt_repo | length > 0) or datadog_remove_custom_repo_file)  and (not ansible_check_mode)
```

That way, we make sure the previous custom repository is removed when:
- `datadog_apt_repo` is not set (that way, we don't keep a previous custom repo activated when we want the official repos)
- `datadog_apt_repo` has been modified (that way, we only use the custom repository currently specified in the role config).

WDYT?

KSerrania (commented):
Thanks for the PR! I left a (pretty big) comment discussing how to make the role idempotent while keeping our current guarantees on the role behavior. WDYT?

tasks/pkg-debian.yml
jharley: @KSerrania thanks looking at my PR and for the thoughtful review! I hadn't considered the case that you outlined and agree that your proposed solution seems like the most straightforward way to address things 🚀

KSerrania: @jharley I'll cherry-pick the changes you made here and add my own then, if you don't mind.

jharley: @KSerrania Not at all! Thank you 🙏

jharley:
Closing in favour of #262

PR #255, opened 2020-01-30
Only try stop sysprobe if it is installed

danielsimkus:
Fixes a bug introduced in https://github.com/DataDog/ansible-datadog/pull/249 where if sysprobe wasn't installed, the provision would still try and stop the service and fail as it doesn't exist

### DIFF
--- a/tasks/agent-linux.yml
+++ b/tasks/agent-linux.yml
@@ -94,7 +94,7 @@
     name: datadog-agent-sysprobe
     state: stopped
     enabled: no
-  when: not datadog_skip_running_check and (not datadog_enabled or not datadog_sysprobe_enabled)
+  when: not datadog_skip_running_check and (not datadog_enabled or not datadog_sysprobe_enabled) and ansible_facts.services['datadog-agent-sysprobe'] is defined
 
 - name: Ensure datadog-agent, datadog-agent-process and datadog-agent-trace are not running
   service:
--- a/tasks/agent-linux.yml
+++ b/tasks/agent-linux.yml
@@ -76,11 +76,15 @@
     enabled: yes
   when: not datadog_skip_running_check and datadog_enabled and not ansible_check_mode
 
+- name: set system probe installed
+  set_fact:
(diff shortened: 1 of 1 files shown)


danielsimkus:
> FAILED! => {"changed": false, "msg": "Could not find the requested service datadog-agent-sysprobe: host"}

tasks/agent-linux.yml
```diff
@@ -94,7 +94,7 @@
     name: datadog-agent-sysprobe
     state: stopped
     enabled: no
-  when: not datadog_skip_running_check and (not datadog_enabled or not datadog_sysprobe_enabled)
+  when: not datadog_skip_running_check and (not datadog_enabled or not datadog_sysprobe_enabled) and ansible_facts.services['datadog-agent-sysprobe'] is defined
```
kevinconaway: We'll want to check both type of services as done here: https://github.com/DataDog/ansible-datadog/blob/8db7a6a88cf40435bb78e54a22dabe8d4fd23b47/tasks/agent-linux.yml#L83

Perhaps we can create a fact about whether the service is installed or not (`datadog_sysprobe_installed`?) and use that here and in the `datadog_sysprobe_enabled` fact check.

kevinconaway (changes requested):
Thanks so much for contributing here Daniel.  One small change and then we should be 🚀

tasks/agent-linux.yml
danielsimkus: Yeah good shout, I'll sort that

danielsimkus: https://github.com/DataDog/ansible-datadog/pull/255/commits/5299d1ae4bf2af14474d83d4d19ab33a1e613df9

kevinconaway (approved):
I tested this locally and it LGTM.  Thanks for your contribution!

KSerrania (approved):
LGTM

danielsimkus:
Hi guys, would we be able to release this on Galaxy? Our provisioning is still failing because of this 👍

KSerrania:
Hi @dv9io0o, I'll do a bugfix release today to include this.

PR #259, opened 2020-02-10
Add skip check on sysprobe set_fact tasks

KSerrania:
### What does this PR do?

Adds `when: not datadog_skip_running_check` to the system-probe `set_fact` tasks.

### Motivation

The `datadog_sysprobe_installed` and `datadog_sysprobe_enabled` facts are only used on tasks with the `when: not datadog_skip_running_check` condition.
Fixes the CI (as it needs to skip the running check).

### DIFF
--- a/tasks/agent-linux.yml
+++ b/tasks/agent-linux.yml
@@ -79,12 +79,14 @@
 - name: set system probe installed
   set_fact:
     datadog_sysprobe_installed: "{{ ansible_facts.services['datadog-agent-sysprobe'] is defined or ansible_facts.services['datadog-agent-sysprobe.service'] is defined }}"
+  when: not datadog_skip_running_check
 
 - name: set system probe enabled
   set_fact:
     datadog_sysprobe_enabled: "{{ system_probe_config is defined
       and system_probe_config['enabled']
       and datadog_sysprobe_installed }}"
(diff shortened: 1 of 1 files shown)


albertvaka (approved):

albertvaka (commented):
How did this code get in, if it didn't pass the CI, though?

KSerrania:
Seems like CircleCI does not run on external contributors' branches from forks (I can't find the jobs that would be associated with #255 neither here: https://circleci.com/gh/DataDog/ansible-datadog nor here: https://circleci.com/gh/dv9io0o/ansible-datadog)

kevinconaway:
Won't this break the tasks below that rely on those two facts to exist?  This would preclude the facts from being _defined_ which will break the checks the below.

KSerrania:
> Won't this break the tasks below that rely on those two facts to exist? This would preclude the facts from being defined which will break the checks the below.

@kevinconaway The two tasks that require these facts also have the `when: not datadog_skip_running_check` filter, so they won't try to evaluate these facts when `datadog_skip_running_check` is set to `true` (thanks to lazy evaluation of the conditions). You can check the CI, which runs with `datadog_skip_running_check: true`.

kevinconaway (approved):

PR #262, opened 2020-02-11
[debian] Fix role idempotence

KSerrania:
### What does this PR do?

After the big 4.0 refactor, the debian part of the role was not idempotent anymore (on every run, two tasks would be marked as "changed", even when no changes were made to the config).
This PR fixes that by adding some logic around the custom repository to keep the current behavior while making the role idempotent.

### Motivation

Make role idempotent again on Debian.

### Additional Notes

Improves upon #250, with help from @jharley.

### DIFF
--- a/tasks/pkg-debian.yml
+++ b/tasks/pkg-debian.yml
@@ -23,11 +23,6 @@
     state: present
   when: datadog_apt_key_url_new is defined
 
-- name: Remove previous datadog apt list file
-  file:
-    path: /etc/apt/sources.list.d/ansible_datadog_agent.list
-    state: absent
-
 - name: Ensure Datadog non-https repositories are deprecated
   apt_repository:
     repo: "{{ item }}"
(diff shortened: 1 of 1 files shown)


albertvaka (approved):

jharley:
@KSerrania Looping back on this: I can confirm that the role is now idempotent. Thank you! 🚀

tasks/pkg-debian.yml
```diff
@@ -51,9 +46,31 @@
...
+- name: Check if custom repository file exists
+  stat:
+    path: /etc/apt/sources.list.d/ansible_datadog_custom.list
+  register: datadog_custom_repo_file
+
+- name: Flag custom repository file for deletion if different from current repository config
+  set_fact:
+    datadog_remove_custom_repo_file: "{{ datadog_repo_file_contents != datadog_apt_repo }}"
+  vars:
+    datadog_repo_file_contents: "{{ lookup('file', '/etc/apt/sources.list.d/ansible_datadog_custom.list') }}"
```
basil, line 62: I'm getting the following error when running this change on my local system against my production inventory:

```
TASK [Datadog.datadog : Flag custom repository file for deletion if different from current repository config] ************************************************************************************************************************
task path: Datadog.datadog/tasks/pkg-debian.yml:58
[WARNING]: Unable to find '/etc/apt/sources.list.d/ansible_datadog_custom.list' in expected paths (use -vvvvv to see paths)
fatal: [basil-docker.dcenter.delphix.com]: FAILED! => {"msg": "An unhandled exception occurred while templating '{{ lookup('file', '/etc/apt/sources.list.d/ansible_datadog_custom.list') }}'. Error was a <class 'ansible.errors.AnsibleError'>, original message: An unhandled exception occurred while running the lookup plugin 'file'. Error was a <class 'ansible.errors.AnsibleError'>, original message: could not locate file in lookup: /etc/apt/sources.list.d/ansible_datadog_custom.list"}
```

The `/etc/apt/sources.list.d/ansible_datadog_custom.list` file exists on the system in my production inventory, but it does not exist on my local system.

Looking at [the documentation for lookup plugins](https://docs.ansible.com/ansible/latest/plugins/lookup.html), I see the following:

>Like all templating, these plugins are evaluated on the Ansible control machine, not on the target/remote.

Perhaps we should be using [the slurp module](https://docs.ansible.com/ansible/latest/modules/slurp_module.html) to read the file from the remote node instead?

KSerrania: Hi @basil,

Thanks for the report! Indeed, using a lookup plugin doesn't work as we wanted it to do. I'll try using the `slurp` module instead and do a bugfix release with this if that works.

KSerrania: Here's the fix: #275

basil: Thanks for the quick fix! I just tested this and it works.

PR #268, opened 2020-03-06
Register result when Agent install task is run

### LINKED ISSUE #266: Add `register` to install task
Add `register` to all install tasks so that `post_tasks` can use `when: agent_install.changed`

KSerrania:
### What does this PR do?

Registers the result of the Agent install in the `datadog_agent_install` variable so that it can be used in `post_tasks`.

### Motivation

Allow custom post tasks to know if the Agent installation did install something.
Fixes #266.


albertvaka (approved):
Is splitting in multiple files needed? There isn't any strong reason, but I think I liked it better on single files.

In any case, LGTM.

KSerrania:
> Is splitting in multiple files needed? There isn't any strong reason, but I think I liked it better on single files.

Variable `registering` happens even if the job is skipped, so for example if we had:

```
- name: Install latest datadog-agent package
  ...
  register: datadog_agent_install
  when: datadog_agent_debian_version is not defined

- name: Install pinned datadog-agent package
  ...
  register: datadog_agent_install
  when: datadog_agent_debian_version is defined
```
in the same file, and installed the latest version, we'd have `datadog_agent_install.changed = False` and `datadog_agent_install.skipped = True` since the second `register` would override the first one.

To avoid that, I needed a way to make the unused task not be in the run list at all, which I managed to by putting them in separate files and only including the files when needed.

There's an alternate solution that we can use if you like it better: we register two internal variables (`datadog_agent_install_latest` and `datadog_agent_install_pinned`) and then create a new task which checks these two vars and sets `datadog_agent_install` to the one that was not skipped.

albertvaka:
Your current solution is fine. Thanks for your explanation 🙇

Issue #265, opened 2020-03-05
Role downloads agent when current version is already up to date


marchenm:
On windows, the role isn't idempotent. It will download the agent every single time even when there isn't a change. There should be a preliminary version check before downloading the agent.

KSerrania:
Hi @marchenm, 

Thanks for the report! I made a PR: #269 to add logic to not download the msi if the detected installed version is the same as the set `datadog_agent_version`, which should fix the idempotency problem for pinned installs (for "latest" installs, the problem is more complex since we can't currently know what the latest package will install without downloading it, so this PR doesn't cover it).

marchenm:
https://docs.ansible.com/ansible/latest/modules/github_release_module.html 

This module should be able to give you the latest release version

KSerrania:
Thanks for the suggestion, however while this module could help in some cases, using the github releases won't work everytime, for two main reasons:
- there can be linux-only or windows-only releases of the datadog-agent (especially bugfix releases); if the latest release is a linux-only release, this module would return a version that doesn't exist on Windows,
- moreover, this will only work for Agent 7 installs (as the Agent 6 and Agent 7 releases are published on the same repository).

joeriharleman:
We're using Ansible to automatically update Datadog, and we'd like to do this rather frequently. However, due to the fact that this Ansible role always downloads the installer, the process can take extremely long. This doesn't scale to multiple servers very well, so it'd be nice to see an improvement.

Any chance of this issue being resolved?

albertvaka:
Hi Joeri! Your concern is valid and it is something we should fix, but that we haven't prioritized yet. I'll bump it internally. Note, though, that not pinning a version is not the recommended configuration of the role. If you pin a specific version of the Agent to install (and update it manually when you think it is convenient) then it won't download the MSI again on each run. Also, pinning ensures that you won't get updates at a time when you don't want them, which is the main reason we recommend it.

grizzlyhazard:
Agreed, more idempotency in this role, please.

bkabrda:
@grizzlyhazard hi 👋 do you have any other issues specifically other than the one reported in the root comment here? If so, I'd definitely like to know what they are - feel free to open more issues or note them here. Thanks!

diederich:
Thanks for looking into this! 🙏  And +1 on somehow allowing to run the role, even with "latest", and not reinstalling on each run. 
[...]

PR #269, opened 2020-03-06
[windows] Fix idempotence when reinstalling same pinned version

KSerrania:
### What does this PR do?

Fetches the currently installed Agent version to avoid downloading the msi installer whenever possible.
It won't work with rc and beta builds (`Get-WmiObject` doesn't return these), nor it will work with latest (since we currently have no way of knowing which version "latest" will install), but this should cover most common pinned installs use cases.

### Motivation

Make role more idempotent.
Should partially fix this issue: #265.

### DIFF
--- a/tasks/pkg-windows.yml
+++ b/tasks/pkg-windows.yml
@@ -4,48 +4,81 @@
     msg: "The Datadog ansible role does not currently support Agent 5"
   when: datadog_agent_major_version|int == 5
 
+# This is a best-effort solution to avoid redownloading the msi when 
+# It won't work with rc / beta builds, and won't work if the version to install
+# is not pinned (as we don't know what version latest will be before downloading
+# the package).
+- name: Check currently installed Agent version
+  win_shell: |
+    $product_name = "Datadog Agent"
+    $query = "Select Name,IdentifyingNumber,InstallDate,InstallLocation,ProductID,Version FROM Win32_Product where Name like '$product_name%'"
+    $installs = Get-WmiObject -query $query
+
+    if (!$installs -or ($installs.Count -eq 0) -or ($installs.Count -gt 1)) {
+      Write-Host ""
+    } else {
+      $ddmaj, $ddmin, $ddpatch, $ddbuild = $installs.Version.split(".")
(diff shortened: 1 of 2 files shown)


tasks/pkg-windows.yml
```diff
@@ -4,48 +4,81 @@
     msg: "The Datadog ansible role does not currently support Agent 5"
   when: datadog_agent_major_version|int == 5
 
+# This is a best-effort solution to avoid redownloading the msi when 
+# It won't work with rc / beta builds, and won't work if the version to install
+# is not pinned (as we don't know what version latest will be before downloading
+# the package).
+- name: Check currently installed Agent version
+  win_shell: |
+    $product_name = "Datadog Agent"
```
albertvaka, line 13: On Chef we use `agent status` for this: https://github.com/DataDog/chef-datadog/blob/483f88a43569b1efd00e7d5c0cb1e479eb09b1a2/libraries/recipe_helpers.rb#L92

I'm not sure which approach is better, but I think we shouldn't have two.

KSerrania: The Chef approach is slightly better because it would work with rc and beta builds. 

However, contrary to this method, it does have a caveat: it only works if you know where the Agent binary is located at (Chef expects it to be `C:/Program Files/Datadog/Datadog Agent/bin/agent`).

One thing to note is that we currently don't offer the possibility to choose the application directory for Windows install; we do have a `datadog_agent_binary_path_windows` parameter, but it's not advertised in the README (and thus I suppose it's an internal parameter) and only used for the integration command, so we could use this option here to use `agent status` (that way people can still configure it for their use case if needed).

albertvaka: If we have a variable for the Agent path, then I think using that to call the binary is reasonable (assuming we won't change the output of `agent status`).

I'm not sure this is necessarily better than your approach, though. I would still sync with @kbogtob, who wrote the Chef code I linked before, to discuss pros and cons of both approaches.

albertvaka (approved):
No need to block this over the discussion on how to query the Agent version, approving.

PR #271, opened 2020-03-24
Ensure the start mode when starting on Windows

DevKyleS:
Ensure the start mode when starting on Windows, this prevents issues when the service was disabled

### DIFF
--- a/tasks/agent-win.yml
+++ b/tasks/agent-win.yml
@@ -41,6 +41,7 @@
   win_service:
     name: datadogagent
     state: started
+    start_mode: auto
     enabled: yes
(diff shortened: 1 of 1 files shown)


DevKyleS:
Could I get a review on this?
@KSerrania

KSerrania (approved):
Hi @DevKyleS,

Thanks for the pull request! The change looks good to me.

This means we'll start the service after each Ansible run, even if the user deliberately disabled the service. I think this behavior is fine (it fits the meaning of the `datadog_enabled` option better, and replicates what we're doing on Linux), but it could break some use cases relying on this specific (faulty) behavior. However, in that case it's easy to fix: they simply have to set `datadog_enabled` to false.

PR #291, opened 2020-08-13
Add support for 3rd party integrations

albertvaka:
Adds support to install 3rd party integrations using the `datadog-agent integration` command.

I also tested that regular (1st party) integrations can still be installed by adding the following to a manual test:
```
    datadog_integration:
      datadog-snmp:
        action: 'install'
        version: '3.7.1'
```

Unfortunately I can't commit the above, since the version used might become superseded by the one shipped in the Agent, making the install fail. That shouldn't be the case for the 3rd party integration, because we won't bundle them, so I left the test there.

### DIFF
--- a/tasks/integration.yml
+++ b/tasks/integration.yml
@@ -49,14 +49,14 @@
 # Install integrations
 
 - name: Install pinned version of integrations (Unix)
-  command: "{{ datadog_agent_binary_path }} integration install {% if 'third_party' in item.value and item.value.third_party %}--third-party{% endif %} {{ item.key }}=={{ item.value.version }}"
+  command: "{{ datadog_agent_binary_path }} integration install {% if 'third_party' in item.value and item.value.third_party == true %}--third-party{% endif %} {{ item.key }}=={{ item.value.version }}"
   become: yes
   become_user: "{{ integration_command_user }}"
   loop: "{{ datadog_integration|dict2items }}"
   when: item.value.action == "install" and ansible_os_family != "Windows"
 
 - name: Install pinned version of integrations (Windows)
-  win_command: "\"{{ datadog_agent_binary_path }}\" integration install {% if 'third_party' in item.value and item.value.third_party %}--third-party{% endif %} {{ item.key }}=={{ item.value.version }}"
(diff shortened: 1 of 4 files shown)


kayayarai (approved):
👍  from docs

tasks/integration.yml
```diff
@@ -49,19 +49,14 @@
 # Install integrations
 
 - name: Install pinned version of integrations (Unix)
-  command:
-    argv:
-      - "{{ datadog_agent_binary_path }}"
-      - integration
-      - install
-      - "{{ item.key }}=={{ item.value.version }}"
+  command: "{{ datadog_agent_binary_path }} integration install {% if 'third_party' in item.value and item.value.third_party %}--third-party{% endif %} {{ item.key }}=={{ item.value.version }}"
```
KSerrania: In Saltstack, I'm doing a strict comparison for `item.value.third_party`: https://github.com/DataDog/datadog-formula/blob/4a82ce48cc95b7acada75633e7227811798e2a53/datadog/config.sls#L64

because otherwise any truthy value works (eg. any non-empty string). Is that also the case here?

albertvaka: That's also the case here, I can make it explicit.

albertvaka: By the way, I didn't find a way to define intermediate variables, so the condition is in the middle of the string. If there's a better way to write this in Ansible let me know :)

KSerrania: You could maybe do something like:

```yaml
- name: Install pinned version of integrations (Unix)
  command: "{{ datadog_agent_binary_path }} integration install (( third_party }}{{ item.key }}=={{ item.value.version }}"
  become: yes
  vars:
    third_party: "{% if 'third_party' in item.value and item.value.third_party == true %}--third-party {% endif %}"
  when: item.value.action == "install" and ansible_os_family != "Windows"
```

We're already using this syntax here: https://github.com/DataDog/ansible-datadog/blob/e08729a7f06868bdd912cdb9bbae7344872a7bf6/tasks/parse-version.yml#L2-L6

However, that assumes that `vars` works correctly with `loop` (ie. that `item` is correctly defined in `vars`), which I haven't tested.

albertvaka: It works and I find it slightly more readable, so I've changed it 👍

KSerrania (approved):

PR #295, opened 2020-09-04
[windows] Add support for installing to different binary directory

derekwbrown:
add support for installing to different config directory

Currently, the MSI installer allows installing to a different binary directory (using `PROJECTLOCATION`) and a different config file root (using `APPLICATIONDATADIRECTORY`) with install command line options.

While those options could be passed to the installer, the portion of the role which sets the agent config did not take that into account, and therefore couldn't modify the config.

With this PR, add two new variable options for specifying the binary directory and config directory.
The role will take care of setting up the install time options _and_ set the directories such that the configuration is properly applied.

PR has additional verification

- On upgrade, ansible role will fail if the user-supplied configuration directory or user-supplied binary directory are different than the original installation directory.

- on upgrade, if there is no user-supplied configuration directory or user-supplied binary directory, the role will query the registry to find the correct configuration; otherwise, on upgrade, any configuration changes written by ansible would not be written to the same directory.

### DIFF
--- a/tasks/pkg-windows-opts.yml
+++ b/tasks/pkg-windows-opts.yml
@@ -11,7 +11,7 @@
 # check the registry. On upgrade, the location of the config file root will 
 # be set here.  
 - name: Check existing config file Directory
-  ansible.windows.win_reg_stat:
+  win_reg_stat:
     path: HKLM:\SOFTWARE\Datadog\Datadog Agent
     name: ConfigRoot
   register: config_root_from_registry
@@ -20,7 +20,7 @@
 # be set here.  
 
 - name: Check existing installPath Directory
-  ansible.windows.win_reg_stat:
+  win_reg_stat:
     path: HKLM:\SOFTWARE\Datadog\Datadog Agent
     name: InstallPath
   register: install_path_from_registry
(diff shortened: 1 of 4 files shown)


tasks/pkg-windows-opts.yml
```diff
@@ -8,6 +8,21 @@
     win_install_args: "{{ win_install_args }} DDAGENTUSER_PASSWORD={{ datadog_windows_ddagentuser_password }}"
   when: datadog_windows_ddagentuser_password | length > 0
 
+- name: Set Program Files Target Directory
+  set_fact:
+    win_install_args: "{{ win_install_args }} PROJECTLOCATION=\"{{ datadog_windows_program_files_dir }}\" "
```
KSerrania, line 65: Nit: I don't think you need the final space, the next argument (if set) will add one.

tasks/pkg-windows-opts.yml
```diff
@@ -8,6 +8,21 @@
...
+  when: datadog_windows_program_files_dir | length > 0
+
+- name: Set Config Files Target Directory
+  set_fact:
+    win_install_args: "{{ win_install_args }} APPLICATIONDATADIRECTORY=\"{{ datadog_windows_config_files_dir }}\" "
```
KSerrania, line 70: Ditto

KSerrania (approved):

KSerrania (approved):
LGTM. Tested the following scenarios:

Install with custom directory -> Upgrade with same custom directory (ok)
Install with custom directory -> Upgrade with different custom directory (fails on upgrade, expected)
Install with default directory -> Upgrade with custom directory (fails on upgrade, expected)
Install with custom directory -> Upgrade with default directory (ok, expected because in the default case we use the vars from the registry).
Install with custom directory -> Rerun with same custom directory (ok, no changed steps so the role is still idempotent).

PR #301, opened 2020-10-01
[CentOS] Use dnf instead of yum with Python 3

### LINKED ISSUE #298: yum module is not supported with Python 3
Hi!

[Documentation](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/yum_module.html) of Ansible's `yum` module says:
> This module only works on Python 2. If you require Python 3 support see the dnf module.

but `ansible-datadog` uses `yum` unconditionally, see https://github.com/DataDog/ansible-datadog/blob/cc9179510369fd2fccdac3403f73394cbd06e0ef/tasks/pkg-redhat/install-pinned.yml#L3 and https://github.com/DataDog/ansible-datadog/blob/cc9179510369fd2fccdac3403f73394cbd06e0ef/tasks/pkg-redhat/install-latest.yml#L3

Python 2 was [retired](https://www.python.org/doc/sunset-python-2/) at January 1, 2020 and I'm working on switching to Python 3 as probably many others.

I believe `ansible-datadog` should use `dnf` module instead of `yum` if the current interpreter is 3.

KSerrania:
### What does this PR do?

Use the `dnf` task instead of `yum` when we detect that a Python 3 interpreter is used on a target host.
Add CentOS 8 tests to the CI.
Add Python 2 / Python 3 tests to the CI.

### Motivation

The `yum` Python module is not compatible with Python 3.
Should fix #298.

### Additional Notes

CentOS / RHEL < 8 + Python 3 interpreter is an invalid setup (you can't use the `dnf` module either because the `python3-dnf` package doesn't exist on these versions of CentOS), so we fail early in that case, with a helpful error message.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -164,7 +164,7 @@ workflows:
               os: ["centos8"]
               python: ["python3"]
 
-      # Newer suse images only have Pythpn 3 installed
+      # Newer suse images only have Python 3 installed
       - test_install:
           matrix:
             parameters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,13 +6,18 @@ commands:
     parameters:
       version:
         type: string
+      python:
+        type: string
     steps:
-      - run: ansible-playbook -i ./ci_test/inventory/ci.ini "./ci_test/install_agent_<<parameters.version>>.yaml" --check
-      - run: ansible-playbook -i ./ci_test/inventory/ci.ini "./ci_test/downgrade_to_5.yaml" --check
+      - run: ansible-playbook -i ./ci_test/inventory/ci.ini "./ci_test/install_agent_<<parameters.version>>.yaml" -e 'ansible_python_interpreter=/usr/bin/<<parameters.python>>' --check
+      - run: ansible-playbook -i ./ci_test/inventory/ci.ini "./ci_test/downgrade_to_5.yaml" -e 'ansible_python_interpreter=/usr/bin/<<parameters.python>>' --check
 
   install_agent_5:
+    parameters:
+      python:
+        type: string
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -90,28 +112,63 @@ jobs:
...
+      # Newer CentOS images only have Pythpn 3 installed
+      - test_install:
+          matrix:
+            parameters:
+              ansible_version: ["2_8", "2_9", "2_10"]
+              agent_version: ["6", "7"]
+              os: ["centos8"]
+              python: ["python3"]
 
+      # Newer suse images only have Pythpn 3 installed
```
mx-psi:
```suggestion
      # Newer suse images only have Python 3 installed
```

mx-psi (approved):

KSerrania:
I'm not 100% sure that we should fail hard when we detect Python 3 + CentOS < 8. On one side, it quickly gives a helpful message and avoids going through half of the role tasks only to find out that it doesn't work, but on the other side, there could be a case where someone backports `dnf` in their CentOS 7 box, in which case the role could succeed.

mx-psi:
> I'm not 100% sure that we should fail hard when we detect Python 3 + CentOS < 8. On one side, it quickly gives a helpful message and avoids going through half of the role tasks only to find out that it doesn't work, but on the other side, there could be a case where someone backports dnf in their CentOS 7 box, in which case the role could succeed.

We could add a flag to ignore the hard fail to deal with that edge case. I don't think it's a likely situation but it doesn't harm to have that flag I suppose.

mx-psi (approved):

apigirl (approved):
👍 docs look good

PR #303, opened 2020-10-01
Add default values when variables are defined but nil

KSerrania:
### What does this PR do?

Adds guards when trying to read variable lengths, to avoid reading the length of `null`, which results in a crash. 

### Motivation

It is possible to get into situations where a variable is set, but has a `null` value (eg. by adding `key: ` in the yaml vars). If we check the length of such a variable (which should thus be a list, string or dictionary), we should first default to a reasonable value if a null value is given.

`default` needs to be used with `boolean=True` to also match falsey values (such as `null`), otherwise it only matches undefined variables.

### DIFF
--- a/tasks/pkg-windows-opts.yml
+++ b/tasks/pkg-windows-opts.yml
@@ -1,7 +1,7 @@
 - name: Set DD Username Arg
   set_fact:
     win_install_args: "{{ win_install_args }} DDAGENTUSER_NAME={{ datadog_windows_ddagentuser_name }}"
-  when: datadog_windows_ddagentuser_name | | default('', true) | length > 0
+  when: datadog_windows_ddagentuser_name | default('', true) | length > 0
 
 - name: Set DD Password Arg
   set_fact:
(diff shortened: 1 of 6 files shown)


tasks/pkg-windows-opts.yml
```diff
@@ -1,12 +1,12 @@
 - name: Set DD Username Arg
   set_fact:
     win_install_args: "{{ win_install_args }} DDAGENTUSER_NAME={{ datadog_windows_ddagentuser_name }}"
-  when: datadog_windows_ddagentuser_name | length > 0
+  when: datadog_windows_ddagentuser_name | | default('', true) | length > 0
```
julien-lebot: There's two consecutive `|` on this line, is that on purpose ?

KSerrania: That's a typo indeed 😄

julien-lebot (approved):
I'm not familiar with Ansible but those changes LGTM !

derekwbrown (approved):

PR #304, opened 2020-10-02
feat(ansible_facts): replace injected facts 'ansible_' with dictionary 'ansible_facts'

samasc30:
Following [Ansible Documentation on Facts](https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#variables-discovered-from-systems-facts) this MR replace injected variables prefixed with `ansible_` with dictionary `ansible_facts`

Using facts with prefix was changed in Ansible 2.5, and should be disabled by default in future releases. In cases where users choose to force only `ansible_facts` to be used by setting `inject_facts_as_vars` to _False_, using older naming leads to errors when applying the role.


albertvaka:
👋 Thanks for the patch! I've tested this on the oldest Ansible we support (2.6) and it works well even there, so I think we can merge it ✅

Thanks again!

PR #308, opened 2020-10-27
Add installation of new debian and GPG keys

derekwbrown:


tasks/pkg-suse.yml
```diff
@@ -27,30 +27,55 @@
...
   rpm_key:
     key: /tmp/DATADOG_RPM_KEY_E09422B3.public
     state: present
   when: not ansible_check_mode
 
+- block:  # Work around due to SNI check for SLES11
+  - name: Stat if 20200908 key (Expires 2024) RPM key already exists
+    stat:
+      path: /tmp/DATADOG_RPM_KEY_20200908.public
+    register: ddnewkey
```
KSerrania: Is it fine to register `ddnewkey` twice (once for the E09422B3 key, once for 20200908)? If it's being overridden, that should be okay (but we need an SLES 11 box to test).

We can also decide to use another variable name (eg. `ddkey_20200908`) to be sure there won't be any conflict.

tasks/pkg-redhat.yml
```diff
@@ -1,24 +1,36 @@
...
+
 - name: Install Datadog yum repo
   yum_repository:
     name: "ansible_datadog_{{ item.key }}"
     description: Datadog, Inc.
     baseurl: "{{ item.value }}"
     enabled: yes
     gpgcheck: yes
-    gpgkey: "{% if datadog_agent_major_version|int == 7 %}{{ datadog_yum_gpgkey_e09422b3 }}{% else %}{{ datadog_yum_gpgkey }}{% endif %}"
+    gpgkey: "{% if datadog_agent_major_version|int == 7 %}{{ datadog_yum_gpgkey_20200908 }}{% else %}{{ datadog_yum_gpgkey }}{% endif %}"
```
KSerrania, line 33: I don't know which one should be put here TBH. Is there a way to specify multiple keys here? If there is, we should put all keys there. If not, I _guess_ we should still put the currently used key (e09422b3) since that's the key people need right now. The steps above will ensure that the future key is present.

derekwbrown: If I understand correctly (and that's a big if)
- This line is actually not necessary due to the manual import steps above
- Since we're manually importing both keys above, it will work either way
- I chose the most recent key so that when the older key expires it doesn't cause problems

KSerrania (approved):

PR #311, opened 2020-11-26
Do not name the RPM repo file differently depending on the Agent version

albertvaka:
## What does this PR do?

Use `datadog.repo` as the name, to be consistent with other install methods.

## Motivation

We might want the repo file to be installed with the package manager in the future, and we don't want to end up with two of them. I'm still not sure how managing this file with Ansible will interact with the package manager trying to install a newer version of the file (eg: because we added a new GPG key). In case the package manager replaces the file (which I think is not the default behavior, but can be configured that way), Ansible will change it back anyway in the next run 🤷 In any case, that's a problem we will have to solve for all config management tools, so I think this change is still making things better by making all install methods consistent.

### DIFF
--- a/tasks/pkg-redhat.yml
+++ b/tasks/pkg-redhat.yml
@@ -67,7 +67,7 @@
   yum_repository:
     name: "ansible_datadog_{{ item }}.repo"
     state: absent
-  with_items: [ 5, 6, 7 ]
+  with_items: [ 5, 6, 7, "custom" ]
 
 - include_tasks: pkg-redhat/install-pinned.yml
   when: datadog_agent_redhat_version is defined
--- a/tasks/pkg-redhat.yml
+++ b/tasks/pkg-redhat.yml
@@ -23,31 +23,51 @@
     state: present
   when: not ansible_check_mode
 
-- name: Install Datadog yum repo
+- name: Install Datadog Agent 5 yum repo
   yum_repository:
-    name: "ansible_datadog_{{ item.key }}"
+    name: datadog
     description: Datadog, Inc.
-    baseurl: "{{ item.value }}"
+    baseurl: "{{ datadog_agent5_yum_repo }}"
     enabled: yes
     gpgcheck: yes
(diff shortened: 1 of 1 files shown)


tasks/pkg-redhat.yml
```diff
@@ -23,32 +23,52 @@
...
     gpgcheck: yes
     gpgkey: "{{ datadog_yum_gpgkey }}"
-    state: present
   when: (datadog_yum_repo | length > 0) and (not ansible_check_mode)
 
+- name: Remove old yum repo files
+  yum_repository:
+    name: "ansible_datadog_{{ item }}.repo"
+    state: absent
+  with_items: [ 5, 6, 7 ]
```
KSerrania: I think we also want to run this with `custom`, to remove the custom yum repository (if it exists).

KSerrania (commented):
LGTM, left one small fix proposal for custom repos

tasks/pkg-redhat.yml
albertvaka: Done and fixed the tests that were failing.

tasks/pkg-redhat.yml
```diff
@@ -23,32 +23,63 @@
...
+- name: Clean repo metadata if repo changed # noqa 503
+  command: yum clean metadata --disablerepo="*" --enablerepo=datadog
+  ignore_errors: yes # Cleaning the metadata is only needed when downgrading a major version of the Agent, don't fail because of this
+  args:
+    warn: no
+  when: repofile5.changed or repofile6.changed or repofile7.changed or repofilecustom.changed
+
+- name: Remove old yum repo files
+  yum_repository:
+    name: "ansible_datadog_{{ item }}.repo"
```
KSerrania: `name` should be the repository id of the repository we want to remove (ie. `ansible_datadog_5`, `ansible_datadog_6`, `ansible_datadog_7`, `ansible_datadog_custom`), not the name of the file containing the repo.

```suggestion
    name: "ansible_datadog_{{ item }}"
```

Otherwise this does nothing, as there are no repos named eg. `ansible_datadog_7.repo`.

albertvaka: Good catch.

KSerrania (approved):

PR #320, opened 2021-01-08
Fix checks delete

albertvaka:
- Missing `state: absent` in a couple places.
- Ansible on Windows needs `win_basename` and `win_file` instead of `basename` and `file`.
- `pattern` should be `patterns`.
- Made `datadog_disable_default_checks` not delete checks in `datadog_additional_checks`.
- Restart the agent if checks were deleted.
- Make linter happy.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,7 +65,7 @@ jobs:
     steps:
       - checkout
       - run: pip install ansible-lint
-      - run: ansible-lint -v .
+      - run: ansible-lint -x 204 -v . # exclude 204 (line too long)
 
   test_install_downgrade:
     parameters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,7 +65,7 @@ jobs:
     steps:
       - checkout
       - run: pip install ansible-lint
-      - run: ansible-lint -x 204 -v . # exclude 204 (line too long)
+      - run: ansible-lint -v .
 
   test_install_downgrade:
     parameters:
--- a/tasks/agent-linux.yml
+++ b/tasks/agent-linux.yml
@@ -39,6 +39,7 @@
 - name: Delete all default checks
   file:
     path: "/etc/datadog-agent/conf.d/{{ item }}.d/conf.yaml.default"
+    state: absent
   loop: "{{ datadog_conf_directories.files | map(attribute='path') | list | map('basename') | list | map('regex_replace', '^(.*).d$', '\\1') | list }}"
   when: datadog_disable_default_checks
 
--- a/tasks/agent-linux.yml
+++ b/tasks/agent-linux.yml
@@ -42,7 +42,7 @@
     path: "/etc/datadog-agent/conf.d/{{ item }}.d/conf.yaml.default"
     state: absent
   loop: "{{ datadog_conf_directories.files | map(attribute='path') | list | map('basename') | list | map('regex_replace', '^(.*).d$', '\\1') | list }}"
-  when: datadog_disable_default_checks and item not in datadog_tracked_check
+  when: datadog_disable_default_checks and item not in datadog_tracked_checks
 
 - name: Ensure configuration directories are present for each Datadog check
   file:
--- a/tasks/agent-linux.yml
+++ b/tasks/agent-linux.yml
@@ -36,6 +36,7 @@
     state: absent
   loop: "{{ datadog_conf_directories.files | map(attribute='path') | list | map('basename') | list | map('regex_replace', '^(.*).d$', '\\1') | list }}"
   when: datadog_disable_untracked_checks and item not in datadog_tracked_checks
+  notify: restart datadog-agent
 
 - name: Delete all default checks
   file:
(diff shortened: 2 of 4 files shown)


.circleci/config.yml
```diff
@@ -65,7 +65,7 @@ jobs:
     steps:
       - checkout
       - run: pip install ansible-lint
-      - run: ansible-lint -v .
+      - run: ansible-lint -x 204 -v . # exclude 204 (line too long)
```
KSerrania: This will fix the CI job, but we may end up with a lower quality score on ansible-galaxy if we do have 204 errors. If we care about that score, we may want to split the offending line(s) instead.

tasks/agent-win.yml
```diff
@@ -17,13 +18,14 @@
   win_file:
     path: "{{ ansible_facts.env['ProgramData'] }}\\Datadog\\conf.d\\{{ item }}.d\\conf.yaml"
     state: absent
-  loop: "{{ datadog_conf_directories.files | map(attribute='path') | list | map('basename') | list | map('regex_replace', '^(.*).d$', '\\1') | list }}"
+  loop: "{{ datadog_conf_directories.files | map(attribute='path') | list | map('win_basename') | list | map('regex_replace', '^(.*).d$', '\\1') | list }}"
   when: datadog_disable_untracked_checks and item not in datadog_tracked_checks
 
 - name: Delete default checks
   file:
```
KSerrania: This should be:

```suggestion
  win_file:
```

tasks/agent-linux.yml
```diff
@@ -39,6 +40,7 @@
 - name: Delete all default checks
   file:
     path: "/etc/datadog-agent/conf.d/{{ item }}.d/conf.yaml.default"
+    state: absent
   loop: "{{ datadog_conf_directories.files | map(attribute='path') | list | map('basename') | list | map('regex_replace', '^(.*).d$', '\\1') | list }}"
   when: datadog_disable_default_checks
```
KSerrania: There's a small edge case here: if you set a default check (like `cpu`) in `datadog_additional_checks`, then the `/etc/datadog-agent/conf.d/cpu.d/conf.yaml` file doesn't get deleted, but `/etc/datadog-agent/conf.d/cpu.d/conf.yaml.default` does get deleted, because there's no `and item not in datadog_tracked_checks` guard on this line.

I don't know if we want to support this edge case or not.

KSerrania (commented):
Aside from the few fixes in comments, there's one more problem: if the only change between two runs is the removal of some check conf files with the new logic, then the Agent does not get restarted by the role. You have to restart the Agent manually to take the removed conf files into account.

.circleci/config.yml
albertvaka: True, ansible-galaxy thinks we still use teletypes with fixed widths. I will split the line.

tasks/agent-linux.yml
albertvaka: Added the `and`.

tasks/agent-linux.yml
```diff
@@ -39,8 +40,9 @@
 - name: Delete all default checks
   file:
     path: "/etc/datadog-agent/conf.d/{{ item }}.d/conf.yaml.default"
+    state: absent
   loop: "{{ datadog_conf_directories.files | map(attribute='path') | list | map('basename') | list | map('regex_replace', '^(.*).d$', '\\1') | list }}"
-  when: datadog_disable_default_checks
+  when: datadog_disable_default_checks and item not in datadog_tracked_check
```
KSerrania:
```suggestion
  when: datadog_disable_default_checks and item not in datadog_tracked_checks
```

tasks/agent-win.yml
```diff
@@ -17,14 +18,15 @@
...
 
 - name: Delete default checks
-  file:
+  win_file:
     path: "{{ ansible_facts.env['ProgramData'] }}\\Datadog\\conf.d\\{{ item }}.d\\conf.yaml.default"
-  loop: "{{ datadog_conf_directories.files | map(attribute='path') | list | map('basename') | list | map('regex_replace', '^(.*).d$', '\\1') | list }}"
-  when: datadog_disable_default_checks
+    state: absent
+  loop: "{{ datadog_conf_directories.files | map(attribute='path') | list | map('win_basename') | list | map('regex_replace', '^(.*).d$', '\\1') | list }}"
+  when: datadog_disable_default_checks and item not in datadog_tracked_check
```
KSerrania:
```suggestion
  when: datadog_disable_default_checks and item not in datadog_tracked_checks
```

tasks/agent5-linux.yml
```diff
@@ -53,8 +55,9 @@
   file:
     path: "/etc/dd-agent/conf.d/{{ item }}.yaml.default"
     state: absent
-  loop: "{{ datadog_conf_files_default.files | map(attribute='path') | list | map('basename') | list | map('regex_replace', '^(.*).yaml.default$', '\\1') | list }}"
-  when: datadog_disable_default_checks
+  loop: "{{ datadog_conf_files_default.files | map(attribute='path') | list
+             | map('basename') | list | map('regex_replace', '^(.*).yaml.default$', '\\1') | list }}"
+  when: datadog_disable_default_checks and item not in datadog_tracked_check
```
KSerrania:
```suggestion
  when: datadog_disable_default_checks and item not in datadog_tracked_checks
```

KSerrania (approved):
LGTM.

PR #325, opened 2021-02-22
Implement usage of multiple GPG keys in repofiles, use keys.datadoghq.com

bkabrda:
This PR adds multiple entries to the repofiles' `gpgkey` entry in order for both yum/dnf/zypper and rpm to have access to all necessary keys. The special `DATADOG_RPM_KEY_CURRENT.public` entry will always contain the key that is used to sign current metadata as well as newly released packages.


KSerrania (approved):
LGTM

PR #332, opened 2021-03-23
Add network_config defaults

KSerrania:
### What does this PR do?

Set default value for `network_config['enabled']`. Otherwise the role fails when this is not explicitly defined.


KSerrania:
Superseded by #335

PR #326, opened 2021-03-04
Add NPM support for Windows

derekwbrown:
Add NPM support for windows
Updates configuration for `network_config` section of datadog.yaml
Update readme to match.

Issue #330, opened 2021-03-23
System Probe Config fails when enabling network_config


lognarly:
When trying to enable NPM on Linux systems, I am getting a failure when trying to set the following:

network_config:
  enabled: true

Role being called by playbook. Everything else works as needed.

```
- name: Install Datadog Agent - JAVA
  include_role:
    name: datadog.datadog
    tasks_from: main
  vars:
    datadog_config: "{{ datadog_agent_config_java }}"
    system_probe_config:
      sysprobe_socket: /opt/datadog-agent/run/sysprobe.sock
    network_config:
      enabled: true
    datadog_checks: "{{ datadog_java_config }}"
```
Error:
```
FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'dict object' has no attribute 'enabled'\n\nThe error appears to be in '/Users/logan/.ansible/roles/datadog.datadog/tasks/agent-linux.yml': line 95, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: set system probe enabled\n  ^ here\n"}
```

The flag used to be:
```
system_probe_config:
  enabled: true
```

Using ansible [core 2.11.0b3]

bkabrda:
Hi 👋 it seems that the README change introduced in https://github.com/DataDog/ansible-datadog/pull/326 made the docs really confusing. Basically for linux at this point, you should keep

```
  system_probe_config:
      enabled: true
      sysprobe_socket: /opt/datadog-agent/run/sysprobe.sock
```

We'll be reverting PR #326, as there are more issues with it. We'll release 4.7.1 with it reverted and when we fix everything properly, we'll release a new version with Windows NPM support and all related updates.

lognarly:
This contrasts with the DD documentation here. So structuring the system-probe.yaml file as you've done should work the same as what is shown in your docs?

https://docs.datadoghq.com/network_monitoring/performance/setup/?tab=agent#setup

bkabrda:
@lognarly thanks to the PR #326 we actually identified an issue with this. `system_probe_config` would only work with agent < 7.24.0 and for newer agents, `network_config` is required. We will fix this and update the docs accordingly for the upcoming release.

lognarly:
Yeah we are installing the latest, so this is required. Is there a time estimation on when the release will be made to allow us to rollout NPM using the parameters required for agents > 7.24.0?

bkabrda:
[...]

PR #335, opened 2021-03-23
Add NPM support for Windows

KSerrania:
### What does this PR do?

Redo of #326, with fixes to avoid undefined variable problems on Windows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,9 +163,11 @@ The following variables are available for live processes:
 
 #### System probe
 
-The [Network Performance Monitoring][7] (NPM) system probe is configured under the `system_probe_config` variable. Any variables nested underneath are written to the `system-probe.yaml`.
+The system probe is configured under the `network_config` variable. Any variables nested underneath are written to the `system-probe.yaml`.
 
-**Note**: The system probe only works on Linux with the Agent v6+.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -176,8 +178,9 @@ datadog_config:
     scrub_args: true
     custom_sensitive_words: ['consul_token','dd_api_key']
 system_probe_config:
-  enabled: true
```
bkabrda, line 179: So this line was removed based on comment https://github.com/DataDog/ansible-datadog/pull/326#discussion_r587487488, but maybe it would be better to leave it here with a note that it may turn on billed features? Providing this option without `enabled: true` is kind of pointless and misleading, as noted in https://github.com/DataDog/ansible-datadog/issues/330.

KSerrania: From our convo: this should not be set directly in recent versions of the Agent, so the documentation is correct (we'll have to add a caveat for older Agent versions). However, that'll require updating the Linux role to not crash when this option is not there (by adding an `if 'enabled' in system_probe_config`) and to support the `network_config['enabled']: true` option.

Let's update the Linux docs + fix the Linux tasks in a separate PR.

bkabrda (approved):

cswatt (approved):

bkabrda (approved):

bkabrda (approved):

PR #336, opened 2021-03-25
 [Linux] Fix system-probe enablement conditions

KSerrania:
### What does this PR do?

Fixes the logic around starting / stopping the system-probe service, with different paths depending on the Agent version:
- before 6/7.18.0: the `datadog-agent-sysprobe` service is independent from the `datadog-agent` service and is started / stopped / restarted by the role, depending on the `system_probe_config.enabled` variable
- between 6/7.18.0 and 6/7.24.0: the `datadog-agent-sysprobe` service depends on `datadog-agent`, therefore only the datadog-agent service needs to be restarted on configuration changes. The `system_probe_config.enabled` variable still defines that behavior.
- since 6/7.24.1: the `network_config.enabled` variable defines if the service should be started or not. The `system_probe_config.enabled` variable is still supported for backwards compatibility, but is not recommended in the documentation anymore.

Makes sure the role doesn't crash when `system_probe_config` has a `None` or empty value.

### Additional notes

TODO: update documentation on how to enable system-probe modules depending on the Agent version (link to the public docs).

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ The following variables are available for live processes:
 
 #### System probe
 
-The system probe is configured under the `network_config` variable. Any variables nested underneath are written to the `system-probe.yaml`.
+The system probe is configured under the `system_probe_config` variable. Any variables nested underneath are written to the `system-probe.yaml`, in the `system_probe_config` section.
 
 [Network Performance Monitoring][7] (NPM) is configured under the `network_config` variable.  Any variables nested underneath are written to the `system-probe.yaml`, in the `network_config` section.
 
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -183,13 +183,15 @@ network_config:
   enabled: true
 ```
 
-Once modification is complete, follow the steps below:
+**Note**: This configuration works with Agent 6.24.1+ and 7.24.1+. For older Agent versions, refer to [the public documentation][8] on how to enable system-probe.
+
+On Linux, once this modification is complete, follow the steps below if you installed an Agent version older than 6.18.0 or 7.18.0:
```
KSerrania, line 188: Added this because on later versions, this should not be needed since the `datadog-agent-sysprobe` service is tied to the `datadog-agent` service.

Another option would be to automate these steps, but I'm not sure it's worth spending time on it given that this only affects a small number of old versions.

albertvaka (approved):

kayayarai (approved):

bkabrda (approved):

PR #337, opened 2021-03-26
Add a note about the dependency on ansible.windows

### LINKED ISSUE #289: Ubuntu 20.04, Ansible 2.10: "couldn't resolve module/action 'win_service'"
The datadog role works fine with Ansible v2.9 but Ubuntu 20.04 requires Ansible v2.10 due to a bugfix.

```
ERROR! couldn't resolve module/action 'win_service'. This often indicates a misspelling, missing collection, or incorrect module path.

The error appears to be in '/Users/mattw/.ansible/roles/datadog.datadog/handlers/main.yml': line 9, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:


- name: restart datadog-agent-win
  ^ here
```

albertvaka:
- Add a note about the dependency on `ansible.windows` since Ansible 2.10. This is because some tasks have been moved from base Ansible into separate collections [1], like `win_service` which is now in `ansible.windows` [2]. Fixes: #289
- Split Windows handler into its own file, so we don't include anything from `ansible.windows` on non-Windows (so we don't have to add the `ansible.windows` dependency on Linux as well).
- Split "Gather Ansible Facts" task in two files for different Ansible versions, prefixed or unprefixed (for some reason in 2.10 it tried to run Powershell on Linux without this when using the task unprefixed as it was before).

Unfortunately there isn't a way the dependency `ansible.windows` is installed automatically when installing our role [3]. In the future, we might be able to provide our own collection that could have an automatic dependency on `ansible.windows`.

[1] https://docs.ansible.com/ansible/devel/porting_guides/porting_guide_2.10.html
[2] https://galaxy.ansible.com/ansible/windows
[3] https://github.com/ansible/ansible/issues/73347#issuecomment-772062084

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,9 +8,9 @@ The Ansible Datadog role installs and configures the Datadog Agent and integrati
 
 - Requires Ansible v2.6+.
 - Supports most Debian and RHEL-based Linux distributions, and Windows.
-- When using Ansible 2.10+, requires the `ansible.windows` collection to be installed (even on Linux):
+- When using Ansible 2.10+ on Windows, requires the `ansible.windows` collection to be installed:
 
-  ```sh
+  ```shell
   ansible-galaxy collection install ansible.windows
   ```
 
--- a/README.md
+++ b/README.md
@@ -8,7 +8,11 @@ The Ansible Datadog role installs and configures the Datadog Agent and integrati
 
 - Requires Ansible v2.6+.
 - Supports most Debian and RHEL-based Linux distributions, and Windows.
-- When using Ansible 2.10+, requires the `ansible.windows` collection to be installed (even on Linux): `ansible-galaxy collection install ansible.windows`
+- When using Ansible 2.10+, requires the `ansible.windows` collection to be installed (even on Linux):
+
+  ```sh
+  ansible-galaxy collection install ansible.windows
+  ```
 
 ### Installation
 
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@ The Ansible Datadog role installs and configures the Datadog Agent and integrati
 
 - Requires Ansible v2.6+.
 - Supports most Debian and RHEL-based Linux distributions, and Windows.
+- When using Ansible 2.10+, requires the `ansible.windows` collection to be installed (even on Linux): `ansible-galaxy collection install ansible.windows`
 
 ### Installation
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -8,6 +8,7 @@ The Ansible Datadog role installs and configures the Datadog Agent and integrati
 
 - Requires Ansible v2.6+.
 - Supports most Debian and RHEL-based Linux distributions, and Windows.
+- When using Ansible 2.10+, requires the `ansible.windows` collection to be installed (even on Linux): `ansible-galaxy collection install ansible.windows`
```
mx-psi: nit: I would move the command to a code block, I think it is more readable that way

mx-psi (approved):

sarina-dd (approved):

KSerrania:
Could we split this in 3 PRs somehow?
- One to add the `ansible.windows` dependency on Windows (+ the handler file split and the gather facts split)
- Another one to fix linting with Ansible 2.10 (file permissions + broken pre_tasks and post_tasks)
- A last one to update the Windows testing docs

tasks/facts-ansible10.yml
```diff
@@ -0,0 +1,3 @@
+---
+- name: Gather Ansible Facts
+  ansible.builtin.setup:
```
mx-psi: What do we need this for?

albertvaka: This task fills the `ansible_facts` dict, see https://github.com/DataDog/ansible-datadog/pull/304

For some reason, the unprefixed task failed for @KSerrania when he ran it on Linux if `ansible.windows` was installed (🤷), but the prefixed task doesn't work in old Ansible versions... so I put each variant on a separate file and we run one or the other conditionally.

mx-psi (approved):

tasks/facts-ansible10.yml
mx-psi: I see, that's weird but if that's the way it works this seems like the best we can do yeah

bkabrda: Could you add a comment explaining this in the code? This is quite unintuitive and I see myself wondering in a month why we added this.

bkabrda (approved):

tasks/facts-ansible10.yml
albertvaka:
```suggestion
  ansible.builtin.setup: # If the full prefix isn't specified in Ansible 2.10+, we might end up running `ansible.windows.setup` instead.
```

PR #338, opened 2021-03-26
Align windows agent to linux so that service is disabled on when datadog_enabled false

erikhjensen:
This includes several proposed modifications:

- Reorder the services in the stop_services task so that when they are shut down, they don't have dependency issues due to shutdown order of operations
- Disable services rather than stopping them when datadog_enabled is false
- Ensure services are not set to disabled when datadog_enabled is true
- Resolve/compute datadog_tracked_checks within the task sequence as set_fact rather than only with the default vars so that its variable precedence is higher than include_vars.  Also address null/none dictionary during this step.   Propose calling it sanitize-checks.

### DIFF
--- a/tasks/agent-win.yml
+++ b/tasks/agent-win.yml
@@ -61,11 +61,12 @@
   win_service:
     name: "{{ item }}"
     state: stopped
+    start_mode: disabled
   when: not datadog_skip_running_check and not datadog_enabled
   with_list:
-    - datadogagent
-    - datadog-process-agent
     - datadog-trace-agent
+    - datadog-process-agent
+    - datadogagent    
 
 - name: Create installation information file
   template:
--- a/tasks/agent-win.yml
+++ b/tasks/agent-win.yml
@@ -52,7 +52,7 @@
 
 - name: Ensure datadog-trace-agent and datadog-process-agent are not disabled
   win_service:
-    name: "{{item}}"    
+    name: "{{ item }}"
     start_mode: manual
   when: not datadog_skip_running_check and datadog_enabled and not ansible_check_mode
   with_list:
--- a/tasks/agent-win.yml
+++ b/tasks/agent-win.yml
@@ -58,7 +58,7 @@
   with_list:
     - datadog-trace-agent
     - datadog-process-agent
-    
+
 - name: Ensure datadog-agent is running
   win_service:
     name: datadogagent
@@ -66,7 +66,7 @@
     start_mode: auto
   when: not datadog_skip_running_check and datadog_enabled and not ansible_check_mode
 
-- name: Ensure datadog-agent is not running
+- name: Ensure datadog-agent is disabled
   win_service:
     name: "{{ item }}"
     state: stopped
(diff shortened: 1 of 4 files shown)


tasks/sanitize-checks.yml
```diff
@@ -0,0 +1,9 @@
+- name: Defend against defined but null datadog_checks variable
+  set_fact:
+      datadog_checks: "{{datadog_checks | default({})}}"
+  when: datadog_checks is none
```
KSerrania: I'd also guard against empty strings and other oddities by doing this:
```suggestion
  set_fact:
      datadog_checks: "{{ datadog_checks | default({}, true) }}"
```

tasks/sanitize-checks.yml
```diff
@@ -0,0 +1,9 @@
...
+
+- name: Resolve datadog_tracked_checks
+  set_fact:
+    datadog_tracked_checks: "{{ datadog_checks | list + datadog_additional_checks }}"
```
KSerrania: To be extra safe, we could also guard against `datadog_additional_checks` being set to None / empty values:
```suggestion
    datadog_tracked_checks: "{{ datadog_checks | list + datadog_additional_checks | default([], true) }}"
```

And, since we're defining it here, I'd remove the `datadog_tracked_checks` definition from the `defaults/main.yml` file

tasks/agent-win.yml
```diff
@@ -50,6 +50,14 @@
...
   notify: restart datadog-agent-win
 
+- name: Ensure datadog-trace-agent and datadog-process-agent are not disabled
+  win_service:
+    name: "{{item}}"    
+    start_mode: manual
+  when: not datadog_skip_running_check and datadog_enabled and not ansible_check_mode
+  with_list:
+    - datadog-trace-agent
+    - datadog-process-agent
```
KSerrania, line 60: Nit: linting
```suggestion
    - datadog-process-agent

```

KSerrania (commented):
Hi @erikhjensen,

Thanks for the PR! I left a few nits, but overall this looks good to me. I tested your changes on Windows and Linux and they're working fine.

tasks/sanitize-checks.yml
erikhjensen: Excellent, I was not aware of this behaviour regarding true.   Good stuff, more concise.

erikhjensen:
Hi @KSerrania I did not perform a reverse pull request to update my fork w/ the nits.  From your comments about testing I am either assuming the updates are innately benign or, that you did test them.   Is there further action for me?

KSerrania:
Hey @erikhjensen, I think these updates are fine. I'll do one last round of testing tomorrow to double-check everything and then approve & merge the PR.

tasks/agent-win.yml
```diff
@@ -50,6 +50,15 @@
   with_items: "{{ datadog_checks|list }}"
   notify: restart datadog-agent-win
 
+- name: Ensure datadog-trace-agent and datadog-process-agent are not disabled
+  win_service:
+    name: "{{item}}"
```
KSerrania: Made a pass with `ansible-lint`, it complained about this:

```suggestion
    name: "{{ item }}"
```

KSerrania (approved):
LGTM, I'll merge once the linting error is fixed.

PR #340, opened 2021-03-29
Update linter and fix issues found

albertvaka:
Bumps the linter docker image and fixes the following warnings:
- File permissions not being explicit.
- `role_name` and `namespace` missing from metadata.
- Wildcards used in pre/post tasks are not expanded (confirmed manually in 2.6 and 2.10)-> I removed the tasks since they never worked so nobody must be using them.

### DIFF
--- a/tasks/agent5-linux.yml
+++ b/tasks/agent5-linux.yml
@@ -3,6 +3,7 @@
   file:
     dest: /etc/dd-agent
     state: directory
+    mode: 0755
 
 - name: (agent5) Create main Datadog agent configuration file
   template:
@@ -10,6 +11,7 @@
     dest: /etc/dd-agent/datadog.conf
     owner: "{{ datadog_user }}"
     group: "{{ datadog_group }}"
+    mode: 0644
   notify: restart datadog-agent
 
 - name: (agent5) Ensure datadog-agent is running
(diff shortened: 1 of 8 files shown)


KSerrania (commented):
LGTM.

About the file permissions: we're already making the Agent 6/7 main config file not world-readable. Should we do the same for the integrations config files, the config dir itself? And should we do the same for the Agent 5 config too?

tasks/agent5-linux.yml
```diff
@@ -67,12 +69,6 @@
     dest: "/etc/dd-agent/conf.d/{{ item }}.yaml"
     owner: "{{ datadog_user }}"
     group: "{{ datadog_group }}"
+    mode: 0644
```
albertvaka:
```suggestion
    mode: 0644 #FIXME: should have permissions set to only be readable by owner
```

tasks/agent5-linux.yml
```diff
@@ -3,13 +3,15 @@
...
     state: directory
+    mode: 0755
 
 - name: (agent5) Create main Datadog agent configuration file
   template:
     src: datadog.conf.j2
     dest: /etc/dd-agent/datadog.conf
     owner: "{{ datadog_user }}"
     group: "{{ datadog_group }}"
+    mode: 0644
```
albertvaka:
```suggestion
    mode: 0644 #FIXME: should have permissions set to only be readable by owner
```

albertvaka:
I've just added a FIXME for A5, the same way I added a FIXME for Windows in the original changes. Those are breaking changes so they will have to wait until a new major version. The Windows one is a bit more worrying than A5. About the directories themselves, I don't think it's a security issue that they are accessible by anyone.

kayayarai (commented):
No docs review required

KSerrania (approved):

bkabrda (approved):

PR #351, opened 2021-04-22
Improvements for APT keys management

bkabrda:
* By default, get keys from keys.datadoghq.com, not Ubuntu keyserver
* Always add the DATADOG_APT_KEY_CURRENT.public key (contains key used to sign current repodata)
* Add 'signed-by' option to all sources list lines
* On Debian >= 9 and Ubuntu >= 16, only add keys to /usr/share/keyrings/datadog-archive-keyring.gpg
* On older systems, also add the same keyring to /etc/apt/trusted.gpg.d


README.md
```diff
@@ -47,9 +47,9 @@ To deploy the Datadog Agent on hosts, add the Datadog role and your API key to y
...
 | `datadog_config_ex`                        | (Optional) Extra INI sections to go in `/etc/dd-agent/datadog.conf` (Agent v5 only).                                                                                                                                                                                                                      |
-| `datadog_apt_repo`                         | Override the default Datadog `apt` repository.                                                                                                                                                                                                                                                            |
+| `datadog_apt_repo`                         | Override the default Datadog `apt` repository. Make sure to use the `signed-by` option if repodata is signed using Datadog's signing keys: `deb [signed-by=/usr/share/keyrings/datadog-archive-keyring.gpg] https://yourrepo`.                                                                            |
```
apigirl:
```suggestion
| `datadog_apt_repo`                         | Override the default Datadog `apt` repository. Make sure to use the `signed-by` option if repo data is signed using Datadog's signing keys: `deb [signed-by=/usr/share/keyrings/datadog-archive-keyring.gpg] https://yourrepo`.                                                                            |
```

apigirl (commented):
just one small content suggestion

README.md
bkabrda: `repodata` is a technical term here and shouldn't be split, but it's true that in the APT world people tend to use `repository metadata` more than `repodata`, so I'll rather use that.

README.md
```diff
@@ -64,7 +64,6 @@ To deploy the Datadog Agent on hosts, add the Datadog role and your API key to y
...
 | `datadog_zypper_gpgkey_e09422b3_sha256sum` | Override the default checksum of the `datadog_zypper_gpgkey_e09422b3` key.                                                                                                                                                                                                                                |
 | `datadog_agent_allow_downgrade`            | Set to `yes` to allow Agent downgrades on apt-based platforms (use with caution, see `defaults/main.yml` for details). **Note**: On Centos this only works with Ansible 2.4+.                                                                                                                             |
-| `use_apt_backup_keyserver`                 | Set to `true` to use the backup keyserver instead of the default one.                                                                                                                                                                                                                                     |
```
mx-psi, line 67: Is removing this a breaking change?

README.md
```diff
@@ -485,26 +484,6 @@ Alternatively, if your playbook **only runs on Windows hosts**, use the followin
 
 ## Troubleshooting
 
-### Debian stretch
```
mx-psi, line 488: Why was this removed? If it's no longer needed, maybe we should, instead of removing, mention the versions in which this happened and that upgrading should fix it? People might not look into the docs of their playbook version

README.md
```diff
@@ -47,9 +47,9 @@ To deploy the Datadog Agent on hosts, add the Datadog role and your API key to y
...
 | `datadog_apt_cache_valid_time`             | Override the default apt cache expiration time (defaults to 1 hour).                                                                                                                                                                                                                                      |
-| `datadog_apt_key_url_new`                  | Override the default URL to Datadog `apt` key (key ID `382E94DE`; the deprecated `datadog_apt_key_url` variable refers to an expired key that's been removed from the role).                                                                                                                             |
+| `datadog_apt_key_url_new`                  | Override the location from which to obtain Datadog `apt` key (key ID `382E94DE`; the deprecated `datadog_apt_key_url` variable refers to an expired key that's been removed from the role). The URL is expected to be a GPG keyring containing keys `382E94DE` and `F14F620E`.                            |
```
KSerrania: I think we can remove this:
```suggestion
| `datadog_apt_key_url_new`                  | Override the location from which to obtain Datadog `apt` key (the deprecated `datadog_apt_key_url` variable refers to an expired key that's been removed from the role). The URL is expected to be a GPG keyring containing keys `382E94DE` and `F14F620E`.                            |
```
as the expected key fingerprints are already specified in the next sentence.

README.md
KSerrania: Not really, given that we don't use keyservers at all anymore. We could leave the option, but it'd do nothing because it has no purpose. We should write somewhere that this option is now useless though.

KSerrania (commented):
The main thing that's worrying me is that the role cannot be idempotent (since we always perform at least one action, importing the current key, each run). While that does reflect the truth, it may not sit well with customers (we've already had issues open because our role would show as doing a change every run).

However, making the role idempotent seems to be a non-trivial effort: one way to do it could be to import the key, and check if something was really imported or changed in the GPG command output (and if not, mark the task as not `changed`), but I don't know how feasible / easy that is in practice.

README.md
bkabrda: Good point on documenting the removal of this option, I'll do that.

README.md
bkabrda: Yeah, you're right about leaving it there and mentioning versions to which this applied.

bkabrda:
@KSerrania that's a good point about idempotency. I think perhaps this might be doable, I'll try to take a look.

bkabrda:
@KSerrania I think I managed to implement a reasonable solution. Here's the comment that I included in the code:

```
# NOTE: in order to display Ansible's `changed: [hostname]` properly throughout
# tasks in this file, we added `changed_when: false` to a lot of them, even if
# they actually run every time (e.g. importing the CURRENT key). The reason is
# that they operate inside a temporary directory and they don't have a
# permanent effect on the host (nothing will actually change on the host
# whether these tasks run or not) except the last one - the actual import of
# the key to `datadog_apt_usr_share_keyring`.
```

KSerrania (approved):

PR #352, opened 2021-04-28
Don't set repo_gpgcheck=1 by default on RHEL/CentOS 8.1 and on custom repos

bkabrda:
Due to [a bug](https://bugzilla.redhat.com/show_bug.cgi?id=1792506) in dnf, `repo_gpgcheck` doesn't work in non-interactive shells in RHEL/CentOS 8.1. This PR implements logic that disables `repo_gpgcheck` on these platforms. It also disables `repo_gpgcheck` on custom repos.


bkabrda:
Hmm, I'll also need to not set `repo_gpgcheck=true` for zypper custom repos. Please hold review until I finish that.

kayayarai (approved):
docs 👍

bkabrda:
Done, ready for review from my side.

mx-psi (approved):

PR #354, opened 2021-04-29
Add a note about installing marketplace integrations. Fixes #344

### LINKED ISSUE #344: Compatibility with installing Marketplace integrations via Ansible role
I was wondering if there were plans to allow Marketplace integrations to be installed via the ansible role?

bkabrda:


mx-psi (approved):

albertvaka:
There was a section already that mentioned `third_party: true` in the README, but it didn't mention marketplace so it wasn't that great. Maybe we can remove that section now?

bkabrda:
@albertvaka the only mention of third party in the README is just below, in the syntax section, which just says `To install third party integrations, set third_party to true:` with an example. Since marketplace integrations are not the only kind of third party integrations, I think it makes sense to just add a section like this. WDYT?

albertvaka:
👍

albertvaka (approved):

apigirl (approved):
docs look good 👍

PR #359, opened 2021-05-17
Make windows package download behavior consistent with linux

camjay:
Skip download in check mode on windows as linux installs do.
This will reporting a changed task in check mode when agent version is current.

### DIFF
--- a/tasks/pkg-windows.yml
+++ b/tasks/pkg-windows.yml
@@ -4,7 +4,7 @@
     msg: "The Datadog ansible role does not currently support Agent 5"
   when: datadog_agent_major_version|int == 5
 
-# This is a best-effort solution to avoid redownloading the msi when 
+# This is a best-effort solution to avoid redownloading the msi when
 # It won't work with rc / beta builds, and won't work if the version to install
 # is not pinned (as we don't know what version latest will be before downloading
 # the package).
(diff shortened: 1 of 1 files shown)


bkabrda:
Hi! Thanks for sending the PR. It looks good, but I think we'll also need https://github.com/DataDog/ansible-datadog/pull/359/files#diff-bef70ee7f6fc30d9bf9be97dd8eaff36eb544b6e51dd36dc3be66f7e4f07610bR105 (the `Delete temporary msi` task) to have the `not ansible_check_mode` in `when`, because otherwise it tries to access the `download_msi_result` variable which is not available in the check mode.

camjay:
thanks for catching that @bkabrda, i've added that conditional.

tasks/pkg-windows.yml
```diff
@@ -106,4 +106,4 @@
   win_file:
     path: "{{ download_msi_result.dest }}"
     state: absent
-  when: (not datadog_skip_windows_install) and (download_msi_result.status_code == 200)
+  when: (not datadog_skip_windows_install) and (download_msi_result.status_code == 200) and (not ansible_check_mode)
```
bkabrda:
```suggestion
  when: (not datadog_skip_windows_install) and (not ansible_check_mode) and (download_msi_result.status_code == 200)
```

The `(not ansible_check_mode)` condition needs to be present *before* `(download_msi_result.status_code == 200)`. That way, Ansible won't try to access `download_msi_result` variable in check mode, thanks to lazy evaluation of conditions (with the way you did it, this would still fail in check mode, because Ansible would still attempt to access the `download_msi_result` variable).

bkabrda:
LGTM now, merging. Thanks a lot for your contribution!

PR #361, opened 2021-05-17
Remove `indentfirst` in system-probe.yaml.j2

### LINKED ISSUE #360: TASK [Datadog.datadog : Create system-probe configuration file] Failure
Since Jinja2 version 3 is released in May 11, the backward compatible `indentfirst` has been [removed](https://github.com/pallets/jinja/pull/1136/commits/b0015c72d5acbf93b9d99a1ce6167889338db85b#diff-d3eeabaeac0164a4aa81d32a82599f1a5c90aca43eb52117d387ae393e9431f7). This breaks system-probe.yaml.j2 . We are getting:
````
fatal: [packer]: FAILED! => {"changed": false, "msg": "AnsibleError: Unexpected templating type error occurred on (# Managed by Ansible

{% if system_probe_config is defined and system_probe_config|length > 0 -%}
system_probe_config:
{# The "first" option is only supported by jinja 2.10+
  which is not present on older systems (CentOS 7, Debian 8, etc.)
  Using the equivalent option "indentfirst" will work with
  all currently existing jinja 2 versions (as of this comment, up to 2.10).
  TODO: when future versions of Jinja are released, check that
  indentfirst is still supported.
 #}
{% filter indent(width=2, indentfirst=True) %}
{{ system_probe_config | to_nice_yaml }}
{% endfilter %}
{% endif %}
): do_indent() got an unexpected keyword argument 'indentfirst'"}
````

Which corresponds to https://github.com/DataDog/ansible-datadog/issues/229

The `indentfirst` was introduced in https://github.com/DataDog/ansible-datadog/pull/240, and we knew the drop of `indentfirst` was comming https://github.com/DataDog/ansible-datadog/pull/240#issuecomment-726327045

### PR DESCRIPTION
**Pull request by tasktop-teho**
### What does this PR do?

Removes specification of the obsolete indent filter argument `identfirst`.
Fixes #360 

### Motivation

The indentfirst option [no longer supported](https://github.com/pallets/jinja/pull/1136/commits/b0015c72d5acbf93b9d99a1ce6167889338db85b#diff-d3eeabaeac0164a4aa81d32a82599f1a5c90aca43eb52117d387ae393e9431f7) in jinja 3

### DIFF
--- a/templates/system-probe.yaml.j2
+++ b/templates/system-probe.yaml.j2
@@ -2,25 +2,14 @@
 
 {% if system_probe_config is defined and system_probe_config | default({}, true) | length > 0 -%}
 system_probe_config:
-{# The "first" option is only supported by jinja 2.10+
-  which is not present on older systems (CentOS 7, Debian 8, etc.)
-  Using the equivalent option "indentfirst" no longer works in jinja 3.
- #}
-{% filter indent(width=2, first=True) %}
+{% filter indent(width=2, True) %}
 {{ system_probe_config | to_nice_yaml }}
 {% endfilter %}
 {% endif %}
 
 {% if network_config is defined and network_config | default({}, true) | length > 0 -%}
 network_config:
-{# The "first" option is only supported by jinja 2.10+
-  which is not present on older systems (CentOS 7, Debian 8, etc.)
-  Using the equivalent option "indentfirst" is supported for
-  all jinja 2 versions up to v2.10.
-  TODO: when future versions of Jinja are released, check that
-  indentfirst is still supported.
- #}
-{% filter indent(width=2, indentfirst=True) %}
+{% filter indent(width=2, True) %}
 {{ network_config | to_nice_yaml }}
 {% endfilter %}
 {% endif %}
--- a/templates/system-probe.yaml.j2
+++ b/templates/system-probe.yaml.j2
@@ -2,14 +2,22 @@
 
 {% if system_probe_config is defined and system_probe_config | default({}, true) | length > 0 -%}
 system_probe_config:
-{% filter indent(width=2, True) %}
+{# The "first" option in indent() is only supported by jinja 2.10+
+  while the old equivalent option "indentfirst" is removed in jinja 3.
+  Using non-keyword argument in indent() to be backward compatible.
+#}
+{% filter indent(2, True) %}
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by bkabrda** [comment]
Thanks for submitting this PR! The problem is that using `first` would break older installations that have Jinja2 < 2.10. I *think* we could work around this by not using `first` at all and adding an extra empty line that would be considered the first unindented line (which would make all of the actual lines indented):

```
system_probe_config:
{% filter indent(width=2) %}

{{ system_probe_config | to_nice_yaml }}
{% endfilter %}
```

This would result in the first line being blank and unindented, and all of the lines of the `system_probe_config` would be indented by 2 spaces correctly. The unindented blank line doesn't break yaml parsing, so everything should, in theory, work ok. I'll need to test this approach carefully to ensure it works as expected.

tasktop-teho:
Thanks @bkabrda . Your solution looks

tasktop-teho:
@bkabrda That's a nice approach. What about not specifying the keyword `first` or `indentfirst`, and let positional argument takes care of it?
`{% filter indent(width=2, True) %}`

bkabrda:
@tasktop-teho you're right that this also possible (and probably even better), but for it to work you'd need to modify the call to be `indent(2, True)`, because you can't use keyword argument before a non-keyword argument. Doing so would lead to an error like this: `jinja2.exceptions.TemplateSyntaxError: invalid syntax for function call expression`. It would also be awesome if you could include a comment above these calls that would explain that we have to keep using non-keyword args because of backwards compatibility. Thanks!

bkabrda:
Looks good now, merging. Thank you for the contribution!

tasktop-teho:
Thank you for the constructive advice

hoshsadiq:
@bkabrda any ideas when there will be a new release with this change? We're running into the same issue.

basil:
Would it be possible to cut a release with this change?

bkabrda:
Sure, I'll do the pre-release tests today and cut a new release if everything checks out - hopefully today/tomorrow, if I find no issues.

bkabrda:
I've just released 4.10.0 version of the Datadog role. See https://github.com/DataDog/ansible-datadog/releases/tag/4.10.0 for changelog. It's now also available on Ansible Galaxy: https://galaxy.ansible.com/DataDog/datadog

PR #362, opened 2021-05-18
ignore mac / osx github datastore files

camjay:


bkabrda:
Looks good, thanks for the PR 👍

Issue #368, opened 2021-06-17
RPM key updates fail in environments where outbound http is prohibited


rossigee:
Symptom:

```
TASK [datadog.datadog : Download new RPM key (Expires in 2024)] ********************************************************************
fatal: [myhostname]: FAILED! => {"changed": false, "dest": "/tmp/DATADOG_RPM_KEY_20200908.public", "elapsed": 10, "msg": "Request failed: <urlopen error timed out>", "url": "http://s3.amazonaws.com/public-signing-keys/DATADOG_RPM_KEY_FD4BF915.public"}
```

Cause:

This is being run in a secured environment where outbound http traffic is prohibited/firewalled.

Action:

Use 'https' URLs 😛

albertvaka:
Closing as I'm going to release a new version that includes your PR. Thanks for the fix! :)

PR #369, opened 2021-06-17
Update 'http' URLs to 'https' wherever possible.

rossigee:
Resolves issue where RPM key cannot be retrieved in secure environments.

https://github.com/DataDog/ansible-datadog/issues/368

### DIFF
--- a/tasks/pkg-debian.yml
+++ b/tasks/pkg-debian.yml
@@ -56,6 +56,9 @@
     state: "absent"
     update_cache: yes
   with_items:
+    - "deb http://apt.datadoghq.com/ stable main"
+    - "deb http://apt.datadoghq.com/ stable 6"
+    - "deb http://apt.datadoghq.com/ stable 7"
     - "deb https://apt.datadoghq.com/ stable main"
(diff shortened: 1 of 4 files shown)


tasks/pkg-debian.yml
```diff
@@ -56,9 +56,6 @@
     state: "absent"
     update_cache: yes
   with_items:
-    - "deb http://apt.datadoghq.com/ stable main"
```
bkabrda: These three should actually stay, as this task is supposed to ensure that non-https source is *absent*, not present.

bkabrda (changes requested):
Hi 👋 thanks for submitting the PR. This is a good catch, I've missed this while implementing the feature. I requested one change inline, so once that's fixed I can merge.

tasks/pkg-debian.yml
rossigee: :facepalm: i should have spotted that. Patch updated.

bkabrda (approved):
Looks great now, I'm merging. Thanks for the PR!

PR #371, opened 2021-06-25
Skip install on Linux systems when pinned version is already installed

bkabrda:
When users are using pinned version of the agent and that version is already installed, we want to skip the whole repository setup + package installation to speed up the role runs significantly. This was already implemented for Windows, so I just made it more general to cover all supported systems.

### DIFF
--- a/tasks/parse-version.yml
+++ b/tasks/parse-version.yml
@@ -36,8 +36,61 @@
     datadog_agent_major_version: "{{ datadog_major }}"
 
 - name: Set OS-specific versions
+  # NOTE: if changing these, make sure the format correspond with values in datadog_version_finding_cmds below
   set_fact:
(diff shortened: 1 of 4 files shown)


tasks/parse-version.yml
```diff
@@ -36,8 +36,61 @@
...
+- name: Create OS-specific version dict
+  set_fact:
+    datadog_agent_os2version:
+      Debian: "{{ datadog_agent_debian_version }}"
+      RedHat: "{{ datadog_agent_redhat_version }}"
+      Suse: "{{ datadog_agent_suse_version }}"
+      Windows: "{{ datadog_agent_windows_version }}"
+
+- name: Get Linux Agent version
+  shell: "{{ datadog_version_finding_cmds[ansible_facts.os_family] }}"  # noqa 305 - Ansible lint things we could use command, but we need shell because some of the cmds have pipes
```
KSerrania:
```suggestion
  shell: "{{ datadog_version_finding_cmds[ansible_facts.os_family] }}"  # noqa 305 - Ansible lint thinks we could use command, but we need shell because some of the cmds have pipes
```

KSerrania (approved):
LGTM

PR #372, opened 2021-06-30
Install datadog-signing-keys package on Debian/Ubuntu

bkabrda:
This installs the datadog-signing-keys package built from https://github.com/DataDog/datadog-signing-keys/. Like this role, it can construct the APT keyring (which is technically not necessary, because the role does that), but it also contains keyrings and policy files for debsig to verify individual DEB package signatures.

### DIFF
--- a/tasks/pkg-debian.yml
+++ b/tasks/pkg-debian.yml
@@ -117,3 +117,11 @@
 
 - include_tasks: pkg-debian/install-latest.yml
   when: datadog_agent_debian_version is not defined
+
+- name: Install latest datadog-signing-keys package
+  apt:
+    name: datadog-signing-keys
+    state: latest  # noqa 403
+    # we don't use update_cache: yes, as that was just done by the install-pinned/install-latest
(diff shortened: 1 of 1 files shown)


albertvaka (approved):

tasks/pkg-debian.yml
```diff
@@ -117,3 +117,11 @@
 
 - include_tasks: pkg-debian/install-latest.yml
   when: datadog_agent_debian_version is not defined
+
+- name: Install latest datadog-signing-keys package
+  apt:
+    name: datadog-signing-keys
+    state: latest  # noqa 403
```
albertvaka, line 124: The dependency we add to the Agent should not clash with this (eg: we should require a version "greater or equal" than X in the Agent).

bkabrda: Yeah. "latest" will always be the desired state for this package. The agent dependency will only limit the lower bound of the version, which is compatible with "latest".

PR #384, opened 2021-08-26
Check `datadog_checks` to avoid misconfiguration

soar:
A misconfiguration like:

```yaml
roles:
  - { name: datadog.datadog, datadog_checks: _datadog_checks }
```

leads to looping over a string and creating a lot of directories in `/etc/datadog-agent/conf.d/`. I propose to add a check for this.

### DIFF
--- a/tasks/sanitize-checks.yml
+++ b/tasks/sanitize-checks.yml
@@ -5,4 +5,8 @@
 - name: Resolve datadog_tracked_checks
   set_fact:
     datadog_tracked_checks: "{{ datadog_checks | list + datadog_additional_checks | default([], true) }}"
-  
+
+- name: Check variables
+  assert:
+    that:
+      - datadog_checks is mapping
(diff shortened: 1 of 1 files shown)


tasks/sanitize-checks.yml
```diff
@@ -5,4 +5,8 @@
 - name: Resolve datadog_tracked_checks
   set_fact:
     datadog_tracked_checks: "{{ datadog_checks | list + datadog_additional_checks | default([], true) }}"
-  
+
+- name: Check variables
```
bkabrda:
```suggestion
- name: Check that datadog_checks is a mapping
```

bkabrda (commented):
Hey 👋 thanks a lot for the PR. It looks good, I just suggested a small change that would ensure the name of the task is a little bit more descriptive. If that looks good to you, let's get it applied and merged.

PR #386, opened 2021-09-21
Fix README.md "Example configuration" indentation

colinrymer:


bkabrda:
Hi 👋 I believe that the example is correct as-is. Some special variables, like `system_probe_config` are intentionally defined outside of `datadog_config` because they're special/trigger other services to start etc. It's true that technically you *could* define them under `datadog_config`, but that wouldn't do all the extra things. Is there a specific issue you're seeing that led you to submitting this PR?
Thanks!

colinrymer:
Ah, I didn't see it earlier, but just saw where the `network_config` (and the other top level config keys referenced in this PR) setting is being handled in the `system_probe.yaml`  rather than in the `datadog.yaml` as provided in https://github.com/DataDog/datadog-agent/blob/main/pkg/config/config_template.yaml#L1258-L1266.

Given that, I'd concur the example is correct as-is, so going to close this PR.

Thanks!

bkabrda:
Thanks!

PR #388, opened 2021-09-29
Create dependabot.yml

mx-psi:
This will run Dependabot monthly


albertvaka (approved):

mx-psi:
Merging, if we don't like it we can revert :shipit:

Issue #387, opened 2021-09-28
Reported vulnerability on rubygem-addressable


mikecowie-seequent:
Our vulneravilty scanner is picking up this as a vulnerability on a container that contains this role.

```
+-------------+------------------+----------+-------------------+---------------+---------------------------------------+
|   LIBRARY   | VULNERABILITY ID | SEVERITY | INSTALLED VERSION | FIXED VERSION |                 TITLE                 |
+-------------+------------------+----------+-------------------+---------------+---------------------------------------+
| addressable | CVE-2021-32740   | HIGH     | 2.7.0             | 2.8.0         | rubygem-addressable:                  |
|             |                  |          |                   |               | ReDoS in templates                    |
|             |                  |          |                   |               | -->avd.aquasec.com/nvd/cve-2021-32740 |
+-------------+------------------+----------+-------------------+---------------+---------------------------------------+

```

I can justify it as mitigated because this is part of the development/test setup,  not the actual role we use. But we (and others with similar scanners, presumably) would appreciate it being bumped.

bkabrda:
Hi 👋 thanks for the report! All of the gems are actually only used for test setup, so the role doesn't use them at runtime. I'll try to make the update soon anyway so your vulnerability scanner is not raising this.

bkabrda:
So the gems were only used in an obsolete testing setup that we were no longer using. I removed all of it in #394, so your security scanner should no longer complain when we release new version. I'm going to go ahead and close this issue.

PR #394, opened 2021-10-11
Remove unused manual test setup

bkabrda:
Nobody is using the manual test setup and it's triggering security scanners because it's obsolete: https://github.com/DataDog/ansible-datadog/issues/387 - let's remove it.


KSerrania (commented):
While we should remove the unused kitchen setup, I don't think we should remove the `manual_tests` folder, as it's separate from the kitchen setup, works and is still useful to quickly set up a local test environment to test local changes.

CONTRIBUTING.md
```diff
@@ -48,57 +48,6 @@ Rebase your changes on `master` and squash your commits whenever possible. This
 
 To contribute, follow the contributing guidelines above.
 
-### Manual testing
-
-To test the roles provided by this project, you can follow the instructions in the manual tests [readme.md][tests].
```
KSerrania, line 53: I think we could keep this line and make it point to the readme in `manual_tests/`?

KSerrania (approved):
LGTM

PR #396, opened 2021-10-18
Docs: mention limitation enabling NPM on Windows

albertvaka:


bkabrda (approved):

README.md
```diff
@@ -167,13 +167,13 @@ The following variables are available for live processes:
...
 
-#### System probe
+#### System probe and Network Performance Monitoring
 
 The system probe is configured under the `system_probe_config` variable. Any variables nested underneath are written to the `system-probe.yaml`, in the `system_probe_config` section.
 
 [Network Performance Monitoring][7] (NPM) is configured under the `network_config` variable.  Any variables nested underneath are written to the `system-probe.yaml`, in the `network_config` section.
 
-**Note**: The system probe works on Linux with Agent v6+. NPM is supported on Windows with Agent v6.27+ and v7.27+.
+**Note for Windows users**: NPM is supported on Windows with Agent v6.27+ and v7.27+. It ships as an optional component that is only installed if `network_config.enabled` is set to true when the Agent is installed or upgraded. Because of this, existing installations might need to do an uninstall+reinstall of the Agent once to install the NPM component, unless the Agent is upgraded at the same time.
```
sarina-dd:
```suggestion
**Note for Windows users**: NPM is supported on Windows with Agent v6.27+. It ships as an optional component that is only installed if `network_config.enabled` is set to true when the Agent is installed or upgraded. Because of this, existing installations might need to do an uninstall and reinstall of the Agent once to install the NPM component, unless the Agent is upgraded at the same time.
```

sarina-dd (approved):
Small suggestions, otherwise LGTM 👍

README.md
albertvaka: Note that `v6.27+ and v7.27+` is different than `v6.27+`, since for example 7.26  doesn't support it.

PR #403, opened 2021-11-15
add integration installation changed_when reporting

camjay:
check for existing version of integrations, don't report changes if integration(s) install is unchanged
add quotes and re-order modified task params for consistency
fix spelling errors

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -182,5 +182,3 @@ datadog_agent7_yum_repo: "https://yum.datadoghq.com/stable/7/{{ ansible_facts.ar
 datadog_agent5_zypper_repo: "https://yum.datadoghq.com/suse/rpm/{{ ansible_facts.architecture }}"
 datadog_agent6_zypper_repo: "https://yum.datadoghq.com/suse/stable/6/{{ ansible_facts.architecture }}"
 datadog_agent7_zypper_repo: "https://yum.datadoghq.com/suse/stable/7/{{ ansible_facts.architecture }}"
-
-integration_results: '{{ integration_existing.results |  json_query ("[?integration.key==integration_existing.item.value]" ) }}'
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -182,3 +182,9 @@ datadog_agent7_yum_repo: "https://yum.datadoghq.com/stable/7/{{ ansible_facts.ar
 datadog_agent5_zypper_repo: "https://yum.datadoghq.com/suse/rpm/{{ ansible_facts.architecture }}"
 datadog_agent6_zypper_repo: "https://yum.datadoghq.com/suse/stable/6/{{ ansible_facts.architecture }}"
 datadog_agent7_zypper_repo: "https://yum.datadoghq.com/suse/stable/7/{{ ansible_facts.architecture }}"
+
+datadog_integration_query: >-
+  [[?integration.key==integration_existing.item.value]].{
+  stderr: ( stderr | default ( omit ) ),
+  stdout: (stdout | default (omit) ),
+  current_version: (item.value.version default ( omit) )}
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -183,4 +183,4 @@ datadog_agent5_zypper_repo: "https://yum.datadoghq.com/suse/rpm/{{ ansible_facts
 datadog_agent6_zypper_repo: "https://yum.datadoghq.com/suse/stable/6/{{ ansible_facts.architecture }}"
 datadog_agent7_zypper_repo: "https://yum.datadoghq.com/suse/stable/7/{{ ansible_facts.architecture }}"
 
-integration_results: integration_existing.results |  jsonquery ("[?integration.key==integration_existing.item.value]" )
+integration_results: '{{ integration_existing.results |  jsonquery ("[?integration.key==integration_existing.item.value]" ) }}'
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -188,9 +188,3 @@ datadog_integration_query: >-
   stderr: ( stderr | default ( omit ) ),
   stdout: (stdout | default (omit) ),
   current_version: (item.value.version | default ( omit) )}
-
-
-
-    integration_existing.results | jsonquery ([?integration.key==integration_existing.item.value].stderr is defined or
-    ( integration.results[?integration.key==integration_existing.item.value].stdout is defined and
-    integration.results[?integration.key==integration_existing.item.value].stdout is not search ( integration_existing.item.value.version ) )
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -187,4 +187,10 @@ datadog_integration_query: >-
   [[?integration.key==integration_existing.item.value]].{
   stderr: ( stderr | default ( omit ) ),
   stdout: (stdout | default (omit) ),
-  current_version: (item.value.version default ( omit) )}
+  current_version: (item.value.version | default ( omit) )}
+
+
+
+    integration_existing.results | jsonquery ([?integration.key==integration_existing.item.value].stderr is defined or
+    ( integration.results[?integration.key==integration_existing.item.value].stdout is defined and
+    integration.results[?integration.key==integration_existing.item.value].stdout is not search ( integration_existing.item.value.version ) )
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -183,4 +183,4 @@ datadog_agent5_zypper_repo: "https://yum.datadoghq.com/suse/rpm/{{ ansible_facts
 datadog_agent6_zypper_repo: "https://yum.datadoghq.com/suse/stable/6/{{ ansible_facts.architecture }}"
 datadog_agent7_zypper_repo: "https://yum.datadoghq.com/suse/stable/7/{{ ansible_facts.architecture }}"
 
-integration_results: '{{ integration_existing.results |  jsonquery ("[?integration.key==integration_existing.item.value]" ) }}'
+integration_results: '{{ integration_existing.results |  json_query ("[?integration.key==integration_existing.item.value]" ) }}'
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -183,4 +183,4 @@ datadog_agent5_zypper_repo: "https://yum.datadoghq.com/suse/rpm/{{ ansible_facts
 datadog_agent6_zypper_repo: "https://yum.datadoghq.com/suse/stable/6/{{ ansible_facts.architecture }}"
 datadog_agent7_zypper_repo: "https://yum.datadoghq.com/suse/stable/7/{{ ansible_facts.architecture }}"
 
-integration_results: '{{  integration_existing.results |  json_query ("[?integration.key=={{ integration.key }}]" ) | first }}'
+integration_results: '{{  integration_existing.results |  json_query ("[?integration.key==integration.key]" ) | first }}'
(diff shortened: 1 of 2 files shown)


bkabrda:
Hi! Thanks for sending the PR. While this seems to work for now, it might break if the output string changed or became localized. I think it would be better to first run a loop with `agent integration show <name> -q` which will fail (if the integration is not present) or echo current version (if it is present). Then we could only install integrations for which `show` failed or returned a version different than what is set in the playbook.

Does that sound reasonable? I can look into it if you don't have time to modify this PR.

camjay:
@bkabrda I've modified it based on your feedback.  Let me know if the spelling fixes should be split out into a separate PR.

defaults/main.yml
```diff
@@ -182,3 +182,5 @@ datadog_agent7_yum_repo: "https://yum.datadoghq.com/stable/7/{{ ansible_facts.ar
 datadog_agent5_zypper_repo: "https://yum.datadoghq.com/suse/rpm/{{ ansible_facts.architecture }}"
 datadog_agent6_zypper_repo: "https://yum.datadoghq.com/suse/stable/6/{{ ansible_facts.architecture }}"
 datadog_agent7_zypper_repo: "https://yum.datadoghq.com/suse/stable/7/{{ ansible_facts.architecture }}"
+
+integration_results: '{{  integration_existing.results |  json_query ("[?integration.key==integration.key]" ) | first }}'
```
bkabrda, line 186: I think this isn't necessary at all (at least I got your PR working and excluded this, see below).

tasks/integration.yml
```diff
@@ -46,22 +46,91 @@
...
+      - "show"
+      - "{{ integration.key }}"
+      - "-q"
   become: yes
   become_user: "{{ integration_command_user }}"
-  vars:
-    third_party: "{% if 'third_party' in item.value and item.value.third_party | bool %}--third-party{% endif %}"
+  check_mode: no
+  changed_when: >-
+    integration_results.rc != 0 or integration_results.stdout != integration.value.version
```
bkabrda, line 63:
```suggestion
    integration_existing.rc != 0 or integration_existing.stdout != integration.value.version
```

tasks/integration.yml
```diff
@@ -46,22 +46,91 @@
...
+
+  # Ignore command failure if integration is not already installed
+  failed_when: no
+  register: "integration_existing"
+  loop: "{{ datadog_integration | dict2items }}"
+  loop_control:
+    loop_var: 'integration'
+  when:
+    - "integration.value.action | lower == 'install'"
+    - "ansible_facts.os_family != 'Windows'"
```
bkabrda, line 73: I prefer to keep the platform checks as the first condition - while at this case it doesn't matter, there is a place in this file where it does matter and it's good to keep things consistent.

Also, please remove the two blank lines below.

tasks/integration.yml
```diff
@@ -46,22 +46,91 @@
...
+
+
+
+- name: "Check for existing install / version of integrations (Windows)"
+  win_command: '"{{ datadog_agent_binary_path }}" integration show {{ item.key }}'
```
bkabrda, line 78:
```suggestion
  win_command: '"{{ datadog_agent_binary_path }}" integration show {{ integration.key }} -q'
```

tasks/integration.yml
```diff
@@ -46,22 +46,91 @@
...
+  become: yes
+  become_user: "{{ integration_command_user }}"
+  check_mode: no
+  changed_when: >-
+    integration_windows_existing.results.stderr is defined or ( integration_windows_existing.results.stdout is defined
+    and integration_windows_existing.results.stdout is not search ( item.value.version ) )
```
bkabrda, line 84:
```suggestion
    and integration_windows_existing.results.stdout is not search ( integration.value.version ) )
```

tasks/integration.yml
```diff
@@ -46,22 +46,91 @@
...
+
+  # Ignore command failure if integration is not already installed
+  failed_when: no
   loop: "{{ datadog_integration|dict2items }}"
-  when: item.value.action == "install" and ansible_facts.os_family != "Windows"
+  register: "integration_windows_existing"
```
bkabrda, line 89: It would be great to also use `integration` loop var here to keep things consistent with the 'Unix' steps.

tasks/integration.yml
```diff
@@ -46,22 +46,91 @@
...
-  win_command: "\"{{ datadog_agent_binary_path }}\" integration install {{ third_party }} {{ item.key }}=={{ item.value.version }}"
+# Install integrations
+
+- name: "Install pinned version of integrations (Unix)"
+  command:
+    argv:
+      - "{{ datadog_agent_binary_path }}"
+      - "integration"
+      - "install"
+      - "{{ third_party }}"
```
bkabrda, line 106: When `third_party` is empty, this makes the command fail (because it will be passed as an empty string argument). You need to use the one-line version.

tasks/integration.yml
```diff
@@ -46,22 +46,91 @@
...
-    third_party: "{% if 'third_party' in item.value and item.value.third_party | bool %}--third-party{% endif %}"
+    third_party: "{% if 'third_party' in check_result.integration.value and check_result.integration.value.third_party | bool %}--third-party{% endif %}"
+  loop: "{{ integration_existing.results  }}"
+  loop_control:
+    loop_var: 'check_result'
+  register: "integration_unix_install"
+  when:
+    - "check_result.integration.value.action | lower == 'install'"
+    - "check_result.rc != 0 or check_result.stdout is undefined or check_result.stdout != check_result.integration.value.version"
+    - "ansible_facts.os_family != 'Windows'"
```
bkabrda, line 121: Again, let's move the platform check to the top.

tasks/integration.yml
```diff
@@ -46,22 +46,91 @@
...
+
+- name: "Install pinned version of integrations (Windows)"
```
bkabrda, line 123: It'd be much better to do the same as we do for Unix, to keep things consistent. A full example of what I'd do is in the PR review summary.

bkabrda (changes requested):
Hey 👋 I needed to do quite a lot of changes to make this work properly on both Linux and Windows. I left bunch of suggestions/comments inline, but if you don't want to go through the review and do the changes one by one, here's a full diff that I have locally that seems to work:

```diff
diff --git a/defaults/main.yml b/defaults/main.yml
index 30f9f47..29795ff 100644
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -186,5 +186,3 @@ datadog_agent7_yum_repo: "https://yum.datadoghq.com/stable/7/{{ ansible_facts.ar
 datadog_agent5_zypper_repo: "https://yum.datadoghq.com/suse/rpm/{{ ansible_facts.architecture }}"
 datadog_agent6_zypper_repo: "https://yum.datadoghq.com/suse/stable/6/{{ ansible_facts.architecture }}"
 datadog_agent7_zypper_repo: "https://yum.datadoghq.com/suse/stable/7/{{ ansible_facts.architecture }}"
-
-integration_results: '{{  integration_existing.results |  json_query ("[?integration.key==integration.key]" ) | first }}'
diff --git a/tasks/integration.yml b/tasks/integration.yml
index 7eaa6ce..20be86b 100644
--- a/tasks/integration.yml
+++ b/tasks/integration.yml
@@ -60,7 +60,7 @@
   become_user: "{{ integration_command_user }}"
   check_mode: no
   changed_when: >-
-    integration_results.rc != 0 or integration_results.stdout != integration.value.version
+    integration_existing.rc != 0 or integration_existing.stdout != integration.value.version

   # Ignore command failure if integration is not already installed
   failed_when: no
@@ -69,42 +69,32 @@
   loop_control:
     loop_var: 'integration'
   when:
-    - "integration.value.action | lower == 'install'"
     - "ansible_facts.os_family != 'Windows'"
-
-
+    - "integration.value.action | lower == 'install'"

 - name: "Check for existing install / version of integrations (Windows)"
-  win_command: '"{{ datadog_agent_binary_path }}" integration show {{ item.key }}'
+  win_command: '"{{ datadog_agent_binary_path }}" integration show {{ integration.key }} -q'
   become: yes
   become_user: "{{ integration_command_user }}"
   check_mode: no
   changed_when: >-
     integration_windows_existing.results.stderr is defined or ( integration_windows_existing.results.stdout is defined
-    and integration_windows_existing.results.stdout is not search ( item.value.version ) )
+    and integration_windows_existing.results.stdout is not search ( integration.value.version ) )

   # Ignore command failure if integration is not already installed
   failed_when: no
   loop: "{{ datadog_integration|dict2items }}"
+  loop_control:
+    loop_var: 'integration'
   register: "integration_windows_existing"
   when:
     - "ansible_facts.os_family == 'Windows'"
-    - "item.value.action | lower == 'install'"
-
-- name: 'debug integration check'
-  debug:
-    var: 'integration_existing'
+    - "integration.value.action | lower == 'install'"

 # Install integrations

 - name: "Install pinned version of integrations (Unix)"
-  command:
-    argv:
-      - "{{ datadog_agent_binary_path }}"
-      - "integration"
-      - "install"
-      - "{{ third_party }}"
-      - "{{ check_result.integration.key }}=={{ check_result.integration.value.version }}"
+  command: "{{ datadog_agent_binary_path }} integration install {{ third_party }} {{ check_result.integration.key }}=={{ check_resul
t.integration.value.version }}"
   become: yes
   become_user: "{{ integration_command_user }}"
@@ -113,24 +103,26 @@
     third_party: "{% if 'third_party' in check_result.integration.value and check_result.integration.value.third_party | bool %}--th
ird-party{% endif %}"
   loop: "{{ integration_existing.results  }}"
   loop_control:
-    loop_var: 'check_result'
+    loop_var: "check_result"
   register: "integration_unix_install"
   when:
+    - "ansible_facts.os_family != 'Windows'"
     - "check_result.integration.value.action | lower == 'install'"
     - "check_result.rc != 0 or check_result.stdout is undefined or check_result.stdout != check_result.integration.value.version"
-    - "ansible_facts.os_family != 'Windows'"

 - name: "Install pinned version of integrations (Windows)"
-  win_command: '"{{ datadog_agent_binary_path }}" integration install {{ third_party }} {{ item.key }}=={{ item.value.version }}'
+  win_command: '"{{ datadog_agent_binary_path }}" integration install {{ third_party }} {{ check_result.integration.key }}=={{ check
_result.integration.value.version }}'
   become: yes
   become_user: "{{ integration_command_user }}"

-  changed_when: "integration_windows_existing.results.stdout is undefined or integration_windows_existing.results.stdout is not sear
ch ( item.value.version )"
-  loop: "{{ datadog_integration|dict2items }}"
-  register: "integration_windows_install"
+  changed_when: "check_result.stdout is undefined or check_result.stdout is not search ( check_result.integration.value.version )"
   vars:
-    third_party: "{% if 'third_party' in item.value and item.value.third_party | bool %}--third-party{% endif %}"
+    third_party: "{% if 'third_party' in check_result.integration.value and check_result.integration.value.third_party | bool %}--th
ird-party{% endif %}"
+  loop: "{{ integration_windows_existing.results }}"
+  loop_control:
+    loop_var: "check_result"
+  register: "integration_windows_install"
   when:
-    - "item.value.action | lower== 'install'"
     - "ansible_facts.os_family == 'Windows'"
-    - "integration_windows_existing.results.stdout is undefined or integration_windows_existing.results.stdout is not search ( item.
value.version )"
+    - "check_result.integration.value.action | lower == 'install'"
+    - "check_result.stdout is undefined or check_result.stdout is not search ( check_result.integration.value.version )"
```

alopezz:
Closing this as abandoned (with conflicts). Will revisit the need for the feature if requested.

PR #407, opened 2022-01-18
Feature: Add tasks for Linux and Windows to create custom Python checks

### LINKED ISSUE #315: Support for Custom Check scripts
We are able to add custom_check yamls when deploying but we need the python scripts for them.
We would like the option to copy .py files with the same name as the script to checks.d folder.
It is essential that the script files will be copied with the same process because Datadog requires a restart.

aRobinson-R7:
The files for these checks must be included in the parent playbook as the path must be relative to the task, or else an absolute path on the local file-system. 
Requires Agent v6+

Fixes: #315

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,25 @@ To configure a custom check use the configuration below. This creates the corres
           - some_data: true
 ```
 
+##### Custom Python Checks
+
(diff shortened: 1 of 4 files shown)


aRobinson-R7:
1. Let me know if tests need created / updated for checking this - it has been tested on machines I have been using
2. Let me know if you would prefer it done another way

PR #409, opened 2022-01-19
error dict object has no attribute 'system'

stegar123:
Error:  while evaluating conditional (ansible_facts.system == \"Linux\"): 'dict object' has no attribute 'system'
During the execution of a certain windows hosts the option "system" in "ansible_facts" is not defined.


albertvaka (approved):
LGTM

PR #410, opened 2022-01-20
Add datadog_manage_config to disable changing the Agent config files

albertvaka:
Adds a new variable `datadog_manage_config` that defaults to `true` but can be set to `false`. Doing so will make our role not change any of the Agent config files when it runs. This allows using the role to update the Agent if manual modifications have been made to the config files without overwriting them.

### DIFF
--- a/tasks/agent-linux.yml
+++ b/tasks/agent-linux.yml
@@ -31,6 +31,7 @@
     mode: 0640
     owner: "{{ datadog_user }}"
     group: "{{ datadog_group }}"
+  when: datadog_manage_config
   notify: restart datadog-agent
 
 - name: Register all checks directories present in datadog
@@ -47,7 +48,7 @@
     path: "/etc/datadog-agent/conf.d/{{ item }}.d/conf.yaml"
     state: absent
   loop: "{{ datadog_conf_directories.files | map(attribute='path') | list | map('basename') | list | map('regex_replace', '^(.*).d$', '\\1') | list }}"
-  when: datadog_disable_untracked_checks and item not in datadog_tracked_checks
+  when: datadog_manage_config and datadog_disable_untracked_checks and item not in datadog_tracked_checks
   notify: restart datadog-agent
 
 - name: Delete all default checks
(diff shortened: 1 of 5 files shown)


tasks/agent-linux.yml
```diff
@@ -47,15 +48,15 @@
     path: "/etc/datadog-agent/conf.d/{{ item }}.d/conf.yaml"
     state: absent
   loop: "{{ datadog_conf_directories.files | map(attribute='path') | list | map('basename') | list | map('regex_replace', '^(.*).d$', '\\1') | list }}"
-  when: datadog_disable_untracked_checks and item not in datadog_tracked_checks
+  when: datadog_manage_config and datadog_disable_untracked_checks and item not in datadog_tracked_checks
```
KSerrania, line 52: You can also add that condition to the task above, to not compute the list of checks if we don't need it.

tasks/agent-win.yml
```diff
@@ -20,35 +21,38 @@
     path: "{{ ansible_facts.env['ProgramData'] }}\\Datadog\\conf.d\\{{ item }}.d\\conf.yaml"
     state: absent
   loop: "{{ datadog_conf_directories.files | map(attribute='path') | list | map('win_basename') | list | map('regex_replace', '^(.*).d$', '\\1') | list }}"
-  when: datadog_disable_untracked_checks and item not in datadog_tracked_checks
+  when: datadog_manage_config and datadog_disable_untracked_checks and item not in datadog_tracked_checks
```
KSerrania, line 24: You can also add that condition to the task above, to not compute the list of checks if we don't need it.

tasks/agent5-linux.yml
```diff
@@ -51,7 +52,7 @@
     path: "/etc/dd-agent/conf.d/{{ item }}.yaml"
     state: absent
   loop: "{{ datadog_conf_files.files | map(attribute='path') | list | map('basename') | list | map('regex_replace', '^(.*).yaml$', '\\1') | list }}"
-  when: datadog_disable_untracked_checks and item not in datadog_tracked_checks
+  when: datadog_manage_config and datadog_disable_untracked_checks and item not in datadog_tracked_checks
```
KSerrania, line 56: You can also add that condition to the two tasks above, to not compute the list of checks if we don't need it.

KSerrania (commented):
Can we add ~unit~ install tests for the `datadog_manage_config: false` case, as it significantly changes the behavior of the role? At least something that verifies that the role doesn't crash in that mode.

albertvaka:
We don't have any unit tests on this repo 😅 😅 😅 Only kitchen tests and a setup for manual tests (I used the later). Should I maybe add a kitchen test?

KSerrania (approved):

PR #412, opened 2022-01-21
Fix provisioning on Python 3 / Amazon Linux 2

### LINKED ISSUE #406: Agent installation fails on aws amzn2
Hi,

we are using the datadog.datadog role from ansible galaxy now for a couple of time. 
Since the last agent version 7.32.4 (how it looks like) our packer build is failing with the following error message:

```
TASK [datadog.datadog : Fail early if Python 3 is used on CentOS / RHEL < 8] ***
    amz-ami: fatal: [default]: FAILED! => {"changed": false, "msg": "The installation of the Agent on CentOS / RHEL versions < 8 requires the 'yum' module, which is not compatible with Python 3. To run this role, use a Python 2 interpreter on hosts running CentOS / RHEL < 8."}
```

This error happen on the latest aws ami `amzn2-ami-hvm-2.0.20211201.0-x86_64-ebs` (`ami-01826408c6676f97e`) and also on the previous version (`ami-0d455065dfd31a3cd`) which we already successfully used in the past.

As a workaround we tried successfully to install the agent by the official shell installer on the same ec2 instance: 
`DD_AGENT_MAJOR_VERSION=7 DD_API_KEY=<KEy> DD_SITE="datadoghq.eu" bash -c "$(curl -L https://s3.amazonaws.com/dd-agent/scripts/install_script.sh)"`

I also tried the workaround described in your docs under https://docs.datadoghq.com/agent/basic_agent_usage/ansible/ which also was not working in our case because the`dnf` and the `python3-dnf` packages are not available in the default aws repositories. If I use the option`datadog_ignore_old_centos_python3_error: true` i get the following error message:

```
TASK [datadog.datadog : Install latest datadog-agent package (dnf)] ************
    amz-ami: fatal: [default]: FAILED! => {"changed": false, "msg": "Could not import the dnf python module using /usr/bin/python3.7 (3.7.10 (default, Jun  3 2021, 00:02:01) [GCC 7.3.1 20180712 (Red Hat 7.3.1-13)]). Please install `python3-dnf` or `python2-dnf` package or ensure you have specified the correct ansible_python_interpreter. (attempted ['/usr/libexec/platform-python', '/usr/bin/python3', '/usr/bin/python2', '/usr/bin/python'])", "results": []}
```

Any ideas or suggestions to handle this issue or why this issue is popping up now?

Here my used version:
- ansible [core 2.12.1]
- packer 1.7.8

datadog.datadog role version:
- 4.12.0
- 4.11.0

aws amis:
- "Name": "amzn2-ami-hvm-2.0.20211201.0-x86_64-ebs",
   "ImageId": "ami-01826408c6676f97e",
   "CreationDate": "2021-12-01T19:55:07.000Z"

- "Name": "amzn2-ami-hvm-2.0.20211005.0-x86_64-ebs",
   "ImageId": "ami-0d455065dfd31a3cd",
   "CreationDate": "2021-10-12T14:12:46.000Z"

moleskin-smile:
### Fix provisioning on Python 3 / Amazon Linux 2.

This (c1db357d65b0c4445d53cb46ffee69ca98bd67d8) is the essence of this PR.

Don't use `python.version.major >= 3` condition to select `dnf` instead
of `yum`, because it breaks Python 3 on Amazon Linux 2 setups, where
`yum` is still used. Use `ansible_pkg_mgr` fact directly instead.

~Remove an explicit fail that's too defensive.~ Improve the explicit "fail early" task to take into account all the variables.

`Vagrantfile` to reproduce the problem:

```
Vagrant.configure("2") do |config|
  config.vm.define "amazonlinux" do |c|
    c.vm.box = "bento/amazonlinux-2"
  end

  # existence of Python 3 is enough to fail this role
  config.vm.provision "shell", inline: "yum install -y python3 python3-pip"

  config.vm.provision "ansible" do |ansible|
    ansible.playbook = "test_7_full.yml"
  end
end
```

Fixes: #406.

### Add manual testing for Centos 8 and Amazon Linux 2

(cb9d25b47d287a180aceedef7840238a0be3102d)

Add manual testing on Centos 8 and Amazon Linux 2

1. Extend manual tests setup to include more operating systems – modern
   RedHat-style one (Centos 8) that uses dnf, legacy RedHat-style one
   which uses yum (Amazon Linux 2). Keep Ubuntu as a default one.
2. Run Ansible on provisioned machines automatically. By default all
   playbooks are run. Now there is no need to change current working
   directory to interact with Ansible and Vagrant. There's also no need
   to modify inventory file if ports change. There're also no problems
   with SSH known hosts anymore.
3. Update `readme.md`.
4. Git ignore `.vagrant`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 *.retry
 .venv
 
+# Vagrant
+.vagrant
+
 # pre and post tasks folders (user defined)
 pre_tasks/
 post_tasks/
--- a/README.md
+++ b/README.md
@@ -548,6 +548,16 @@ On Debian Stretch, the `apt_key` module used by the role requires an additional 
     datadog_api_key: "<YOUR_DD_API_KEY>"
 ```
 
+### CentOS 6/7 with Python 3 interpreter and Ansible 2.10.x or below
+
+The `yum` Python module, which is used in this role to install the Agent on CentOS-based hosts, is only available on Python 2 if Ansible 2.10.x or below is used. In such cases, the `dnf` package manager would have to be used instead.
+
+However, `dnf` and the `dnf` Python module are not installed by default on CentOS-based hosts before CentOS 8. In this case, it is not possible to install the Agent when a Python 3 interpreter is used.
+
+This role fails early when this situation is detected to indicate that Ansible 2.11+ or a Python 2 interpreter is needed when installing the Agent on CentOS / RHEL < 8.
+
+To bypass this early failure detection (for instance, if `dnf` and the `python3-dnf` package are available on your host), set the `datadog_ignore_old_centos_python3_error` variable to `true`.
+
 ### Windows
 
 Due to a critical bug in Agent versions `6.14.0` and `6.14.1` on Windows, installation of these versions is blocked (starting with version `3.3.0` of this role).
--- a/README.md
+++ b/README.md
@@ -548,14 +548,6 @@ On Debian Stretch, the `apt_key` module used by the role requires an additional 
     datadog_api_key: "<YOUR_DD_API_KEY>"
 ```
 
-### CentOS 6/7 with Python 3 interpreter
-
-The `yum` Python module, which is used in this role to install the Agent on CentOS-based hosts, is only available on Python 2. When a Python 3 interpreter is detected on a target host, the `dnf` package manager and the `dnf` Python module are used instead.
-
-However, `dnf` and the `dnf` Python module are not installed by default on CentOS-based hosts before CentOS 8. In this case, it is not possible to install the Agent when a Python 3 interpreter is used. This role fails early when this situation is detected to indicate that a Python 2 interpreter is needed when installing the Agent on CentOS / RHEL < 8.
-
-To bypass this early failure detection (for instance, if `dnf` and the `python3-dnf` package are available on your host), set the `datadog_ignore_old_centos_python3_error` variable to `true`.
-
 ### Windows
 
 Due to a critical bug in Agent versions `6.14.0` and `6.14.1` on Windows, installation of these versions is blocked (starting with version `3.3.0` of this role).
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -85,6 +85,13 @@ datadog_yum_gpgkey_20200908: "https://s3.amazonaws.com/public-signing-keys/DATAD
 datadog_yum_gpgkey_20200908_sha256sum: "4d16c598d3635086762bd086074140d947370077607db6d6395b8523d5c23a7d"
 # Default zypper repo and keys
 
+# By default, we fail early & print a helpful message if an older Ansible version and Python 3
+# interpreter is used on CentOS < 8. The 'yum' module is only available on Python 2, and the 'python3-dnf'
+# package is not available before CentOS 8.
+# If set to true, this option removes this check and allows the install to proceed. Useful in specific setups
+# where an old CentOS host using a Python 3 interpreter does have 'dnf' (eg. through backports).
+datadog_ignore_old_centos_python3_error: false
+
 # By default, the role uses the official zypper Datadog repository for the chosen major version
 # Use the datadog_zypper_repo variable to override the repository used.
 datadog_zypper_repo: ""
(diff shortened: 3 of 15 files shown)


bkabrda:
Hey 👋 thanks for sending the PR. Overall it looks great, but I would like to keep the condition `Fail early if Python 3 is used on CentOS / RHEL < 8`. If this is a problem for you on Amazon Linux (which I'm assuming it is), could you add another part to the condition, something like `ansible_facts.distribution != 'Amazon'`? That should make it work for Amazon Linux, but also keep the condition working for CentOS/RHEL. Thanks!

moleskin-smile:
Hey! Thanks a lot for looking at this.

Regarding the `Fail early` condition – could you please explain what is the point of failing here? Python 3 can still be used on Centos 7 without problems. In my opinion using `python.version.major >= 3` condition to determine package manager is not correct regardless of Linux distribution.

The following setup works with my fixes, but breaks if we restore `Fail early` or use `dnf` based on Python version:

`❯ head -5 test_7_full.yml`
```
- hosts: all
  roles:
    - { role: ../../ansible-datadog, become: yes }  # On Ansible < 1.9, use `sudo: yes` instead of `become: yes`
  vars:
    ansible_python_interpreter: /usr/bin/python3
```
`❯ cat Vagrantfile`
```
Vagrant.configure("2") do |config|
  config.vm.define "centos" do |c|
    c.vm.box = "centos/7"
  end

  config.vm.provision "shell", inline: "yum install -y python3 python3-pip"

  config.vm.provision "ansible" do |ansible|
    ansible.playbook = "test_7_full.yml"
  end
end
```

cswatt (approved):
looks good for docs!

bkabrda:
Hey 👋 are you using Ansible 4 or newer? Ansible 4 (which uses ansible-core >= 2.11.0) implemented `respawn_module` [1] (which gets used by `yum` module) which will respawn the python process running on the host under Python 2 and thus your play would pass even without the condition, while it would fail on ansible < 4.

Because we still support Ansible >= 2.6, we want to keep this condition present (and I'll be more than happy to drop it once we only support Ansible >= 4).

Does this make sense? If so, could you please re-add the condition and also add the `Amazon` clause to it, so it would work for you? Thanks! 

[1] https://github.com/ansible/ansible/commit/4c5ce5a1a9e79a845aff4978cfeb72a0d4ecf7d6

moleskin-smile:
Thanks, that explains a lot! So I've re-added the `Fail early` task, but altered it the way that's more specific. I think this way it won't break real setups but still would be helpful directing users towards possible solutions.

So the new logic to fail is the following:

```
  when: (not datadog_ignore_old_centos_python3_error)
        and (ansible_version.full is version("2.11", operator="lt", strict=True))
        and (ansible_pkg_mgr == "yum")
        and (ansible_facts.python.version.major | int >= 3)
```

I've tested this manually with `test_7_full.yml` playbook and:

1. Ansible 2.8.20 + Python 3 + Centos/7 – early fail
2. Ansible 2.8.20 + Python 3 + Amazonlinux – early fail
3. Ansible 2.8.20 + Python 3 + Centos/8 – pass
4. Ansible 2.8.20 + default Python + Centos/7 – pass
5. Ansible 2.8.20 + default Python + Amazonlinux – pass
6. Ansible 2.8.20 + default Python + Centos/8 – pass
7. Ansible core 2.12.1 + Python 3 + Centos/7 – pass
8. Ansible core 2.12.1 + Python 3 + Amazonlinux – pass
9. Ansible core 2.12.1 + Python 3 + Centos/8 – pass
10. Ansible core 2.12.1 + default Python + Centos/7 – pass
11. Ansible core 2.12.1 + default Python + Amazonlinux – pass
12. Ansible core 2.12.1 + default Python + Centos/8 – pass

BTW, can we also test with Ansible 2.11+ on CircleCI?

bkabrda (approved):
Looks great, merging.

BTW I'm working on adding ansible 2.11 tests in CI and will hopefully have them set up later this week. I tested this PR locally with ansible 2.11 and it works fine.

daroga0002:
@bkabrda when is planned release of module with this feature?

bkabrda:
@daroga0002 I'd like to do a new release this week. I'm currently doing pre-release testing, so we'll see if I hit any problems or not. I'll update this issue later today with the progress.

bkabrda:
@daroga0002 I've just released version 4.14.0 with this feature: https://github.com/DataDog/ansible-datadog/releases/tag/4.14.0 Enjoy!

daroga0002:
Thank you

PR #414, opened 2022-01-25
Support jinja2_native=true setting

### LINKED ISSUE #288: Set Agent default major version: object of type 'int' has no len()
Using the following task:
```yml
- name: Install DataDog
  include_role:
    name: datadog.datadog
  vars:
    datadog_api_key: ${datadog_api_key}
```
```
TASK [datadog.datadog : Convert datadog_agent_major_version to string] ******************************************************************************************************************************************
ok: [XX.XXX.XXX.XX]

TASK [datadog.datadog : include_tasks] **************************************************************************************************************************************************************************
skipping: [XX.XXX.XXX.XX]

TASK [datadog.datadog : Set Agent default major version] ********************************************************************************************************************************************************
fatal: [XX.XXX.XXX.XX]: FAILED! => {"msg": "The conditional check 'datadog_agent_major_version | length == 0' failed. The error was: Unexpected templating type error occurred on ({% if datadog_agent_major_version | length == 0 %} True {% else %} False {% endif %}): object of type 'int' has no len()\n\nThe error appears to be in '/home/XXX/.ansible/roles/datadog.datadog/tasks/set-parse-version.yml': line 9, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Set Agent default major version\n  ^ here\n"}
```

https://github.com/DataDog/ansible-datadog/blob/3787f2ee24267cc8605f26f6521715ab3ece57df/tasks/set-parse-version.yml#L2-L12
It seems like the conversion int -> string in this file doesn't work as intended.

Workaround is: `    datadog_agent_major_version: "7"`

bkabrda:
This is an attempt to fix #288 by ensuring that we handle Jinja native types properly. It also adds one matrix test to run with `ANSIBLE_JINJA2_NATIVE=true` to ensure that the basic functionality works ok with this setting.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,8 +27,11 @@ commands:
         type: string
       python:
         type: string
+      jinja2_native:
+        type: string
+        default: "false"
     steps:
-      - run: ansible-playbook -i ./ci_test/inventory/ci.ini "./ci_test/install_agent_<<parameters.version>>.yaml" -e 'ansible_python_interpreter=/usr/bin/<<parameters.python>>'
+      - run: ANSIBLE_JINJA2_NATIVE="<<parameters.jinja2_native>>" ansible-playbook -i ./ci_test/inventory/ci.ini "./ci_test/install_agent_<<parameters.version>>.yaml" -e 'ansible_python_interpreter=/usr/bin/<<parameters.python>>'
       - run: datadog-agent version
 
   test_install_no_manage_config:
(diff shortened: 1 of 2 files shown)


albertvaka (approved):
Do you understand why this breaks with `jinja2_native`? The PR passes the tests but it's quite mysterious to me why it broke and why this fixes it 🤔

bkabrda:
@albertvaka the setting turns on the `nativetypes` feature of Jinja: https://jinja.palletsprojects.com/en/2.10.x/nativetypes/. This basically means that anything that looks like an int (e.g. `7` for `datadog_major`) will be an int, even if it's quoted in the yaml file.

PR #416, opened 2022-01-26
Move Windows version task to separate file to prevent dependency on ansible.windows

### LINKED ISSUE #413: Role doesn't work on an Ubuntu 20 machine because of missing `ansible.windows` dependency
I'm running the role from a mac machine targetting a ubuntu 20 machine. It seems that this line is the offending one as it's not skipped for some reason.

```
# NOTE: This won't work with rc / beta builds.
- name: Get Windows Agent version
  win_shell: |
    $product_name = "Datadog Agent"
    $query = "Select Name,IdentifyingNumber,InstallDate,InstallLocation,ProductID,Version FROM Win32_Product where Name like '$product_name%'"
    $installs = Get-WmiObject -query $query

    if (!$installs -or ($installs.Count -eq 0) -or ($installs.Count -gt 1)) {
      Write-Host ""
    } else {
      $ddmaj, $ddmin, $ddpatch, $ddbuild = $installs.Version.split(".")
      Write-Host "$($ddmaj).$($ddmin).$($ddpatch)"
    }
  register: datadog_version_check_win
  changed_when: false
  failed_when: false
  check_mode: no
  when: ansible_facts.os_family == "Windows"
```
Here is the error log 

```
TASK [datadog.datadog : Initialize skip install flag to false] *****************************************************
Tuesday 25 January 2022  13:01:34 +0300 (0:00:00.078)       0:00:08.527 ******* 
ok: [site-20220125T084204]

TASK [datadog.datadog : include_tasks] *****************************************************************************
Tuesday 25 January 2022  13:01:34 +0300 (0:00:00.053)       0:00:08.580 ******* 
fatal: [site-20220125T084204]: FAILED! => {"reason": "couldn't resolve module/action 'win_shell'. This often indicates a 
misspelling, missing collection, or incorrect module path.\n\nThe error appears to be in 
'/Users/abdellah/.ansible/roles/datadog.datadog/tasks/parse-version.yml': line 70, column 3, but may\nbe elsewhere in the file 
depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n# NOTE: 
This won't work with rc / beta builds.\n- name: Get Windows Agent version\n  ^ here\n"}
```

I ended up installing `ansible.windows` to move on.

bkabrda:
Fixes #413. Extracting the `win_shell` task to a separate file that gets included conditionally prevents ansible from failing on parsing the task when `ansible.windows` collection is not installed.

### DIFF
--- a/tasks/parse-version-windows.yml
+++ b/tasks/parse-version-windows.yml
@@ -1,0 +1,18 @@
+# NOTE: This won't work with rc / beta builds.
+- name: Get Windows Agent version
+  win_shell: |
+    $product_name = "Datadog Agent"
+    $query = "Select Name,IdentifyingNumber,InstallDate,InstallLocation,ProductID,Version FROM Win32_Product where Name like '$product_name%'"
+    $installs = Get-WmiObject -query $query
+
+    if (!$installs -or ($installs.Count -eq 0) -or ($installs.Count -gt 1)) {
+      Write-Host ""
+    } else {
+      $ddmaj, $ddmin, $ddpatch, $ddbuild = $installs.Version.split(".")
+      Write-Host "$($ddmaj).$($ddmin).$($ddpatch)"
+    }
+  register: datadog_version_check_win
+  changed_when: false
+  failed_when: false
+  check_mode: no
+  when: ansible_facts.os_family == "Windows"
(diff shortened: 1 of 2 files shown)


albertvaka (approved):
✅

PR #418, opened 2022-02-02
Support Rocky and AlmaLinux, migrate CI tests from EOL'd CentOS 8 to Rocky Linux 8

bkabrda:
Let's support Rocky Linux and AlmaLinux. For the CI part, CentOS 8 was EOL'd on 31st January 2022. We still need to test on a RHEL-compatible system, so let's switch to Rocky Linux.


albertvaka (approved):
LGTM

PR #426, opened 2022-03-11
Add support for installing Datadog Agent on macOS

ghost:
This code adds tasks to support installing Datadog Agent on macOS

* only supports version 6 and 7 as this is the versions supported by the offical `https://s3.amazonaws.com/dd-agent/scripts/install_mac_os.sh` install script
* follows the general process of the official install script but omits an 'upgrade' option that converted datadog.conf to datadog.yaml


lc-applause:
This is still a work in progress, tracking down all relevant locations that need changes.

README.md
```diff
@@ -7,7 +7,7 @@ The Ansible Datadog role installs and configures the Datadog Agent and integrati
 ### Requirements
 
 - Requires Ansible v2.6+.
-- Supports most Debian and RHEL-based Linux distributions, and Windows.
+- Supports most Debian and RHEL-based Linux distributions, macOS and Windows.
```
kayayarai:
```suggestion
- Supports most Debian and RHEL-based Linux distributions, macOS. and Windows.
```

README.md
```diff
@@ -299,6 +301,17 @@ When the variable `datadog_windows_download_url` is not set, the official Window
...
+#### macOS
+
+When the variable `datadog_macos_download_url` is not set, the official macOS DMG package corresponding to the `datadog_agent_major_version` is used:
+
+| # | Default macOS DMG package URL                                |
+|---|--------------------------------------------------------------|
+| 6 | https://s3.amazonaws.com/dd-agent/datadogagent.dmg           |
+| 7 | https://s3.amazonaws.com/dd-agent/datadog-agent-7-latest.dmg |
+
+To override the default behavior, set this variable to something else than an empty string.
```
kayayarai:
```suggestion
To override the default behavior, set this variable to something other than an empty string.
```

README.md
```diff
@@ -299,6 +301,17 @@ When the variable `datadog_windows_download_url` is not set, the official Window
 
 To override the default behavior, set this variable to something else than an empty string.
 
+#### macOS
+
+When the variable `datadog_macos_download_url` is not set, the official macOS DMG package corresponding to the `datadog_agent_major_version` is used:
+
+| # | Default macOS DMG package URL                                |
```
kayayarai:
```suggestion
| Agent version | Default macOS DMG package URL                                |
```

kayayarai (approved):
A couple minor edit suggestions, but otherwise, the documentation looks good.

README.md
ghost: updated the column header

README.md
ghost: updated wording

README.md
ghost: added punctuation

ci_test/install_agent_7_macos.yaml
```diff
@@ -0,0 +1,29 @@
+---
+
+- hosts: all
+  roles:
+    - { role: '/Users/distiller/project/' }
+  vars:
+    datadog_api_key: "11111111111111111111111111111111"
+    datadog_user: "distiller"
```
ghost: I am slightly concerned about the `datadog_user` property. It is used extensively in the linux and windows code to set the owner of datadog files, etc. The default `dd-agent` value is of course not available on the circleci macs or my orgs machines so without setting this `datadog_user` property the install does not function correctly. Maybe I am misunderstanding something? I am worried about user error if this property is not specified.

ghost: Maybe the intention is that outside of this playbook the user must create the `dd-agent` user or specify an override. I don't believe it should be the responsibility of this playbook to create users as that could have significant consequences on the running machines.

albertvaka: Hi @lc-applause,

First of all, thanks a ton for writing this patch 🙇 It's great and we will try to allocate time soon to review and test it.

About the `datadog_user` property: while it's true that this is a modifiable property, it's not documented anywhere and there's no way the role can work if it is changed even on Linux and Windows, so feel free to ignore any edge cases related to it. Checking the history I see it was added at least 6 years ago, when probably things were quite different. At some point we should remove it.

Albert

ghost: we at Applause are beginning to deploy our macs as well as linux (debian-based) machines in a development environment using this branch - I've notices a concerning issue around the `datadog_user` property. The linux `apt` tasks require `become: yes` - since i didn't want to edit the code related to apt in this repo I set `become: yes` at the top level when adding this role to my main playbook. This causes problems with macOS, specifically the `launchd` tasks that query `launchctl` to verify the agent service is running. The macOS Datadog agent installer creates a `.plist` file for the ssh session user regardless of `become`. The ultimate problem is that since the `.plist` file is in the directory `/Users/<name>/Library/LaunchAgents` I need to add extra `become_user:` to the `launchd` tasks.
I hope that makes sense?
I will try to make it work with ignoring the `datadog_user` property.

ghost:
I am not sure why the lint task started failing - seems circleci node is having trouble installing ansible-lint package. I will double check if it is something I changed

ci_test/install_agent_7_macos.yaml
ghost: The easiest solution I can think of is to move the `.plist` file created by the Datadog Agent installer to the root folder `/Library/LaunchAgents/` - but I am wondering if that would be acceptable to the Datadog team.

ghost: i was able to resolve these issues by not using `community.general.launchd` and calling launchctl commands directly

tasks/integration.yml
```diff
@@ -46,6 +56,18 @@
   loop: "{{ datadog_integration|dict2items }}"
   when: item.value.action == "remove" and ansible_facts.os_family == "Windows"
 
+- name: Removing integrations (macOS)
```
albertvaka: This is the same as on Linux, we don't need a macOS-specific job.

ghost: you are right - removed this one and renamed the task above it to make it clear it runs on both unix-like and macOS

bkabrda:
@lc-applause hey 👋 I'm about to start reviewing your PR. Since it's a big one and introduces some larger functionality, it will probably take some time. In the meanwhile, I'd like to ask you to stop force pushing to make the review easier for me. If you still need to fix/improve things, please add more commits to this PR. We can always squash all of them before merging. Thanks!

ci_test/install_agent_7_macos.yaml
bkabrda: Yes, I think doing a systemwide LaunchDaemon installation (== putting plist files in `/Library/LaunchAgents`) is the way to go here. Do note that I actually have a PR under development for our `install_mac_os.sh` script that does just that to enable system-wide installation without the user logged in via gui: https://github.com/DataDog/datadog-agent/pull/10621 - it'd probably make sense to make my PR and your PR behave consistently, so I'll do my review in that spirit.

ghost:
Thank you, I will not force push while you are reviewing. The code here is in a reasonably good state, we have used it to deploy a few new machines. There was a slight issue with Datadog config files not being accessible by the login user which caused them not to load. Other than that we have had few problems.
There is some strangeness when someone has installed Datadog manually though the UI and activated the taskbar utility menu. This seems to be related to activating load-on-start from the taskbar menu interfering with modifying launchctl items via the terminal. I have not fully understood this edge case.

ci_test/install_agent_7_macos.yaml
ghost: I was familiarizing myself with the `launchctl` options and noted that Apple considers many of the available commands deprecated. You will see I chose to use the non-deprecated command forms that include `domain/uid` such as `launchctl bootstrap gui/<uid> <plist-file>`. There are other domains which seem better suited to automation (system, user) but I was not able to get the plist file included with datadog agent to load in those domains. I am not familiar enough with macOS details like this to understand what is happening. Unfortunately the `gui` domain requires the user to have a logged in gui session on the target machine.

bkabrda: My PR (mentioned above) that I did for `install_mac_os.sh` actually does what you're trying to achieve, I think. It moves the plist file from user's homedir to `/Library/LaunchDaemons` and adds a UID/GID entry so that the Agent runs under desired user/group. Have you tried that? Feel free to take some inspiration from my PR might help.

I think it would also make a lot of sense to create `datadog_macos_user`/`datadog_macos_group` variables and run the agent under these to ensure things are configurable.

bkabrda:
Hi 👋 yeah, the code overall looks really good. I'll probably have couple minor comments further down the line, but now I'd like to concentrate on how the agent is actually run - and I think it should be run as a systemwide LaunchDaemon as noted in one of the threads above.

I would be glad to offer my hands-on help with this PR - if you agree on running the agent as systemwide LaunchDaemon and if you're ok with giving me access to your fork, I can do some of the work myself and you can see if it makes sense to you. WDYT?

ghost:
I think it sounds straightforward enough - I will make some changes this morning and validate on our integration environment. We were discussing internally the benefits of moving the plist file to the root `Library` folder, we have a machine that likes to reboot itself and dd agent wont come back online without logging into the machine ATM.

tasks/agent-macos.yml
```diff
@@ -0,0 +1,175 @@
...
+      cmd: "mv /Users/{{ ansible_user }}/{{ datadog_macos_user_plist_file_path }} {{ datadog_macos_system_plist_file_path }}"
+      removes: "/Users/{{ ansible_user }}/{{ datadog_macos_user_plist_file_path }}"
+      creates: "{{ datadog_macos_system_plist_file_path }}"
+    become: true
+
+  - name: Inject user and group data into system plist file
+    shell:
+      cmd: "cp -f {{ datadog_macos_system_plist_file_path }} /tmp/.plist.backup
+            && 
+            awk -v d=\"</dict>\"
```
ghost, line 35: i chose to use `awk` because at Applause we often install `gnu-sed` over the built in `sed` (BSD) and they are a bit different.

tasks/agent-macos.yml
```diff
@@ -0,0 +1,175 @@
+---
+- name: 'Load user data into JSON string {"uid": <number>, "gid": <number>}'
+  shell:
+    cmd: "dscacheutil -q user -a name {{ ansible_user }} | awk 'BEGIN { RS=\"\\n\"; ORS=\" \" } /uid:/ { print \"{ \\\"uid\\\": \" $2\",\" } /gid:/ { print \"\\\"gid\\\": \" $2 \" }\"}'"
+    executable: /bin/bash
+  changed_when: false
+  register: macos_user_output
```
ghost, line 7: i've chosen not to parameterize the user/group - the playbook will use the ssh session user (`ansible_user`)

tasks/agent-macos.yml
```diff
@@ -0,0 +1,175 @@
...
+
+- name: Extract JSON user data as variable object
+  set_fact:
+    macos_user_data: "{{ macos_user_output.stdout }}"
```
ghost, line 15: this is maybe a bit obscure - but since the string inside `macos_user_output.stdout` is valid json (i format it that way) ansible will auto parse and assign the object to the variable

ghost:
> I would be glad to offer my hands-on help with this PR - if you agree on running the agent as systemwide LaunchDaemon and if you're ok with giving me access to your fork, I can do some of the work myself and you can see if it makes sense to you. WDYT?

I don't have permission to grant collaboration access to this repo.
If you want to take over the final development internally you are welcome to grab the code or I can target this PR to merge into some other upstream branch of your choice.

We are actively deploying machines with this code so I have a set of daily test machines and am happy to see this PR through to the end.

tasks/agent-macos.yml
bkabrda: Sounds like a good choice 👍

tasks/agent-macos.yml
bkabrda: While I understand that this way is good for your usecase, it's inconsistent with the rest of the role and might not be good enough for everyone. I think we should really make this configurable.

tasks/agent-macos.yml
bkabrda: Could you add a comment that explains this? I think it's ok to do like this, but deserves the explanation.

bkabrda:
> If you want to take over the final development internally you are welcome to grab the code or I can target this PR to merge into some other upstream branch of your choice.

That sounds really good. I'll create a branch today that you can target and we can merge this PR. I can then make a followup PR with my suggestions/modifications and let you review it. When we're both happy, we can merge that branch into main.

bkabrda:
I created a `macos-support` branch. Please re-target your PR to it and I'll merge so I could start working on my suggested changes.

ghost:
> I created a `macos-support` branch. Please re-target your PR to it and I'll merge so I could start working on my suggested changes.

PR is re-targeted

tasks/agent-macos.yml
ghost: i can't personally imagine the use case, but it is good to have the option.

tasks/agent-macos.yml
ghost: yes, you are right, more detail is needed in context - i've added supporting comments

bkabrda:
Thanks! I'll merge the PR and then work on some minor modifications. I'll open a new PR with my modifications and ask you for a review, so we can agree on a solution that works for you as well as other users of the role.

PR #437, opened 2022-04-28
Add macOS support

bkabrda:
This PR adds macOS support to our ansible role. Some notes:

* DD Agent is configured to run as system-wide LaunchDaemon.
* Due to the way the scripts in our `dmg` package work, the package will install a login item and start a user-specific LaunchAgent if (and only if) `ansible_user` is also currently logged in via GUI. There are tasks that handle this by deactivating the LaunchAgent and login item when this happens.
* I refactored most of the Linux tasks to a separate file to be able to share them with macOS.
* The user under which the DD Agent should run can be specified via `datadog_macos_user` and defaults to `ansible_user`.
* It's possible to override the `dmg` download URL by specifying `datadog_macos_download_url`.

Huge thanks go to @lc-applause for sending the basis of this functionality!


alai97 (approved):
Looks good for docs!

bkabrda:
I added another commit to fix ansible-lint issues. For some reason, ansible-lint started complaining on files that weren't touched by this PR. Technically it would've been better if I did another PR to fix the issues, but I figured that this would be easier and avoid rebasing.

README.md
```diff
@@ -292,12 +296,23 @@ Since version 4.9.0, the `use_apt_backup_keyserver` variable has been removed, a
...
-To override the default behavior, set this variable to something else than an empty string.
+To override the default behavior, set this variable to something other than an empty string.
+
+#### macOS
+
+When the variable `datadog_macos_download_url` is not set, the official macOS DMG package corresponding to the `datadog_agent_major_version` is used:
+
+| Agent version | Default macOS DMG package URL                                |
+|---------------|--------------------------------------------------------------|
+| 6             | https://s3.amazonaws.com/dd-agent/datadogagent.dmg           |
```
albertvaka:
```suggestion
| 6             | https://s3.amazonaws.com/dd-agent/datadog-agent-6-latest.dmg |
```
Both are the same DMG, but the code actually uses this one.

tasks/agent-macos.yml
```diff
@@ -0,0 +1,92 @@
+---
+# NOTE: the DMG gets installed as ansible_user, but we then configure it to run
+# under datadog_macos_user and remove the user-specific config for ansible_user
+- name: Load user data
+  shell:
+    cmd: "dscacheutil -q user -a name {{ datadog_macos_user }} | awk 'BEGIN { RS=\"\\n\"; ORS=\" \" } /uid:/ { print \"{ \\\"uid\\\": \" $2\",\" } /gid:/ { print \"\\\"gid\\\": \" $2 \" }\"}'"
```
albertvaka, line 6: 🙄

tasks/agent-macos.yml
```diff
@@ -0,0 +1,92 @@
...
+- name: Set permissions for DataDog Directories
+  file:
+    path: "{{ item }}"
+    owner: "{{ macos_user_data.uid }}"
+    group: "{{ macos_user_data.gid }}"
+    recurse: yes
+  with_items:
+    - "datadog_macos_etc_dir"
+    - "datadog_macos_logs_dir"
+    - "datadog_macos_run_dir"
```
albertvaka:
```suggestion
    - "{{ datadog_macos_etc_dir }}"
    - "{{ datadog_macos_logs_dir }}"
    - "{{ datadog_macos_run_dir }}"
```
I believe it should be like this.

albertvaka (approved):
Left some comments but LGTM.

tasks/agent-macos.yml
bkabrda: You're right! Ansible created these directories (literally `datadog_macos_etc_dir` etc) and set the permissions on them. It did work on my system because all of the users I tested with belonged to the same group. Great catch!

PR #443, opened 2022-06-29
Add includepkgs to repo config

sspans-sbp:
The current repository configuration is unrestricted. Users / patch tooling running the standard `yum update` command will deploy the latest datadog agent. The recommended configuration of the role is to specify an explicit version, resulting in unfortunate upgrade/downgrade actions.

This PR adds the versioned package as `includepkgs` to the repository configuration. 
This ensures that only the configured version can be installed by yum / dnf.

### DIFF
--- a/tasks/pkg-redhat.yml
+++ b/tasks/pkg-redhat.yml
@@ -55,6 +55,16 @@
     key: /tmp/DATADOG_RPM_KEY_20200908.public
     state: present
   when: not ansible_check_mode
+  
+- name: Set versioned includepkgs variable
+  set_fact:
+    datadog_includepkgs: "{{ datadog_agent_flavor }}-{{ datadog_agent_redhat_version | regex_replace('^\\d+:', '') }}"
+  when: datadog_agent_redhat_version is defined
+
+- name: Set plain includepkgs variable
+  set_fact:
+    datadog_includepkgs: "{{ datadog_agent_flavor }}"
+  when: datadog_agent_redhat_version is not defined
 
 - name: Install Datadog Agent 5 yum repo
   yum_repository:
@@ -62,6 +72,7 @@
     description: Datadog, Inc.
     baseurl: "{{ datadog_agent5_yum_repo }}"
     enabled: yes
+    includepkgs: "{{ datadog_includepkgs }}"
     repo_gpgcheck: no  # we don't sign Agent 5 repodata
     gpgcheck: "{{ datadog_yum_gpgcheck }}"
     gpgkey: [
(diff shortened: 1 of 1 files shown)


mx-psi:
Hey @sspans-sbp thanks for your contribution, could you make sure the CI tests pass? It seems like they are failing on downgrades

sspans-sbp:
Includepkgs doesn't use the epoch for matching, I've added a jinja filter to strip it from the version string.

sspans-sbp:
It seems this feature is also supported by zypper, so it makes sense to add it for SuSE as well.

mx-psi (approved):
LGTM, but I want @bkabrda to have a look too

sspans-sbp:
Any progress on the review?

mx-psi:
> Any progress on the review?

Hey, my teammate is out of office, so you will have to wait a few more days until he is back (I will remind him, no worries :))

templates/zypper.repo.j2
```diff
@@ -13,6 +13,7 @@ name=Datadog, Inc.
 enabled=1
 autorefresh=1
 baseurl={{ baseurl }}
+includepkgs={{ datadog_includepkgs }}
```
bkabrda: I've tried testing this on OpenSUSE 42.3 and 15.2 and it works on neither, it seems that it just gets ignored. Does it work for you? Do you have any documentation reference that says it should work with zypper?

bkabrda (commented):
Thanks for opening the PR. Overall I think this is a good thing to do, however as noted inline, I can't get the SUSE part to work (I don't think this actually works with zypper). Once we get that sorted out, I can merge.

templates/zypper.repo.j2
sspans-sbp: You are indeed correct. My testing shows the same, includepkgs is ignored by Zypper.
I'll update the PR.

sspans-sbp:
I've removed the SuSE bits, please review.

bkabrda (approved):
LGTM, merging. Thanks for the PR!

bkabrda:
As a side note, for a while I was deliberating about the fact that some people might actually expect to be able to do plain `yum update` after running the playbook and get the latest version. I think the reasoning in the description of this PR is valid and its pros (ensuring that the agent flavor/version are always as configured) far outweigh the potential downside, especially considering Ansible's design goal of ensuring that the target machine is put to a certain state.

That said, this new behavior should be documented clearly and explicitly in the release notes for the next release so that users are well informed of its consequences.

PR #452, opened 2022-08-05
Feat: add the possibility to prevent the zypper repository installation

jb-mayer:
This can be useful in an environment where the use of public repository is not allowed.


defaults/main.yml
```diff
@@ -100,6 +100,9 @@ datadog_ignore_old_centos_python3_error: false
 # Use the datadog_zypper_repo variable to override the repository used.
 datadog_zypper_repo: ""
 
+# Define if the official zypper Datadog repository services should be installed
+datadog_zypper_repo_enabled: yes
```
bkabrda: Could you rename this to `datadog_manage_zypper_repo`? The `enabled` keyword is a repofile keyword and it might the variable name confusing. Additionally, we already have a variable called `datadog_manage_config`, so using `datadog_manage_zypper_repo` would be consistent with that.

bkabrda (changes requested):
Hi! Thank you very much for submitting this PR! It looks good, but I have a suggestion for changing the naming of the variable. When you fix that, I can merge.

defaults/main.yml
bkabrda: Actually, even better might be `datadog_manage_zypper_repofile`.

jb-mayer: Hi, I changed the variable name according to your suggestion.
Thanks.

bkabrda (approved):
LGTM, merging. Thanks a lot for the contribution!

PR #454, opened 2022-09-02
Use ansible_managed instead of hardcoded message

jlosito:
Makes use of [`{{ ansible_managed }}`](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/template_module.html#synopsis) which defaults to "Ansible managed" instead of custom hardcoded message.


bkabrda (approved):
LGTM, thank you very much for opening the PR. I'm merging.

PR #468, opened 2023-01-13
Fix CI by pinning proper Agent 5 version

bkabrda:
We pin 5.32.9 on RHEL-like systems, but we use 5.32.8 everywhere else.


amenasria (approved):
LGTM, well done !

PR #478, opened 2023-04-03
Redhat : Refresh Datadog repository cache

Pythyu:
Fix a support issue by ensuring that dnf keys are imported into the local db.

### DIFF
--- a/tasks/pkg-redhat.yml
+++ b/tasks/pkg-redhat.yml
@@ -146,7 +146,9 @@
     warn: no
   when: repofile5.changed or repofile6.changed or repofile7.changed or repofilecustom.changed
 
-# On certain version of dnf, gpg keys aren't imported into the local db. This rule assure that they are correctly imported into the local db.
+# On certain version of dnf, gpg keys aren't imported into the local db with the package install task.
+# This rule assures that they are correctly imported into the local db and users won't have to manually accept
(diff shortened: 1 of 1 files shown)


tasks/pkg-redhat.yml
```diff
@@ -146,6 +146,13 @@
     warn: no
   when: repofile5.changed or repofile6.changed or repofile7.changed or repofilecustom.changed
 
+- name: Refresh Datadog repository cache # noqa 503
```
bkabrda, line 152: Could you please explicitly add a comment about the corner case that lead us to adding this task?

tasks/pkg-redhat.yml
```diff
@@ -146,6 +146,14 @@
     warn: no
   when: repofile5.changed or repofile6.changed or repofile7.changed or repofilecustom.changed
 
+# On certain version of dnf, gpg keys aren't imported into the local db. This rule assure that they are correctly imported into the local db.
```
bkabrda:
```suggestion
# On certain version of dnf, gpg keys aren't imported into the local db with the package install task.
# This rule assures that they are correctly imported into the local db and users won't have to manually accept
# them if running dnf commands on the hosts.
```

bkabrda (approved):

PR #479, opened 2023-04-05
[ansible] Ensure user/group selected for macOS systemwide installation actually exist

Pythyu:
Ensure that the user and group associated exists.

### DIFF
--- a/tasks/pkg-macos.yml
+++ b/tasks/pkg-macos.yml
@@ -6,6 +6,26 @@
     msg: "The Datadog ansible role does not currently support Agent 5 on macOS"
   when: datadog_agent_major_version|int == 5
 
+- name: Check if user doesn't exist
+  command: id -u "{{ datadog_macos_user }}" >/dev/null 2>&1
+  register: mac_user_check
+  ignore_errors: true
+
+- name: Fail if user doesn't exist
+  fail:
+    msg: "The Datadog ansible role wasn't able to find the user : {{ datadog_macos_user }}"
+  when: mac_user_check|failed
+
+- name: Check if group doesn't exist
+  command: dscacheutil -q group | grep "name:" | awk '{print $2}' | grep -w "{{ macos_user_group.stdout }}" >/dev/null 2>&1
+  register: mac_group_check
+  ignore_errors: true
+
+- name: Fail if group doesn't exist
+  fail:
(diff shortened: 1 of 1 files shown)


tasks/pkg-macos.yml
```diff
@@ -6,6 +6,35 @@
...
+  register: macos_group_data_gid
+
+- name: Check if group exist
+  shell:
+    cmd: "dscacheutil -q group -a gid {{ macos_group_data_gid }} | grep '^name: ' | awk '{ print $2 }'"
+  register: mac_group_check
+  changed_when: false
+  ignore_errors: true
+
+- name: Fail if group doesn't exist
```
bkabrda: I'm wondering whether this is necessary. Since we don't accept the group name as user input, but taking user's primary group from the system, is there a case where the system could tell us group that doesn't exist?

Pythyu: To be honest, I had the same question. I think it's very unlikely but yea the only case it could be happening to me is if someone disable SIP(Security Integrity Protection) and edit users and group system file by hand. In doubt I kept the check but tbh I don't see any advantage by hand modifying mac os system so we could remove it. What do you think ?

bkabrda: Yes, I think it is not our task to ensure that the system is actually giving us correct information. Let's remove all the group-related tasks here.

tasks/pkg-macos.yml
```diff
@@ -6,6 +6,17 @@
     msg: "The Datadog ansible role does not currently support Agent 5 on macOS"
   when: datadog_agent_major_version|int == 5
 
+- name: Check if user exist
```
bkabrda:
```suggestion
- name: Check if the macOS user for Agent service exists
```

tasks/pkg-macos.yml
```diff
@@ -6,6 +6,17 @@
...
+  command: id -u "{{ datadog_macos_user }}"
+  register: mac_user_check
+  changed_when: false
+  ignore_errors: true
+
+- name: Fail if user doesn't exist
```
bkabrda:
```suggestion
- name: Fail if the macOS user for Agent service doesn't exist
```

bkabrda (changes requested):

bkabrda (approved):

PR #481, opened 2023-05-04
Add support for configuring APM injection

rarguelloF:
This PR adds support for configuring APM injection for a host (see: https://docs.datadoghq.com/tracing/trace_collection/library_injection_local/?tab=host#configure-docker-injection).

The supported scenarios are:
- Agent and applications running on the same host (`datadog_apm_instrumentation_enabled: host`).
- Agent and applications running in separate Docker containers on the same host (`datadog_apm_instrumentation_enabled: docker`).
- All the previous scenarios at the same time (`datadog_apm_instrumentation_enabled: all`).


defaults/main.yml
```diff
@@ -163,6 +163,13 @@ datadog_windows_config_root: "{{ ansible_facts.env['ProgramData'] }}\\Datadog"
 # arguments to supply to the windows installer.
 win_install_args: " "
 
+# A boolean to enable host injection.
+datadog_apm_host_injection_enabled: false
+# A boolean to enable docker injection.
+datadog_apm_docker_injection_enabled: false
+# List of apm libraries to install if host or docker injection is enabled.
+# Valid values are: java, js, dotnet, all
```
bkabrda: I think we have `python` now as well, right?

I think in general this comment should contain a note "as of writing this comment", because the set of libraries might change in time which could be confusing to users with older version of this role.

tasks/apm-host-injection.yml
```diff
@@ -0,0 +1,35 @@
+---
+- name: Check if platform is supported
```
bkabrda: These task names get printed to users, so let's be more descriptive. I'd suggest something like `Check if APM injection is supported on this host OS`

tasks/apm-host-injection.yml
```diff
@@ -0,0 +1,35 @@
...
+  fail:
+    msg: APM Injection is not supported in this platform
+  when: ansible_facts.os_family not in ["Debian", "RedHat", "Rocky", "AlmaLinux"]
+
+- name: Flush handlers
+  meta: flush_handlers
+
+- name: Ensure datadog-agent is running
```
bkabrda: Why does the Agent need to be running at this point?

If indeed it does need to be running, it should have `when: not ansible_check_mode`. You can learn more about the check mode here: https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_checkmode.html#using-check-mode

If it doesn't need to be running at this point, we can probably also remove the `flush_handlers` task.

tasks/apm-host-injection.yml
```diff
@@ -0,0 +1,35 @@
...
+  include_tasks: pkg-redhat/install-apm-host-inject-latest.yml
+  when: ansible_facts.os_family in ["RedHat", "Rocky", "AlmaLinux"]
+
+# TODO: enable when flag is ready
+#- name: Check if dd-host-install needs to be run
+#  command: dd-host-install --dry-run
+#  register: dd_host_install_cmd_result
+#  ignore_errors: true
+
+- name: Run dd-host-install
```
bkabrda: Couple points here:
* Should this be run conditionally only when `datadog_apm_host_injection_enabled` is set to `true`?
* Are we missing the analogous call for `datadog_apm_docker_injection_enabled`?
* Can we make the name more descriptive for users, e.g. `Run APM host injection setup script`?
* Again, this should get `when: not ansible_check_mode`.

tasks/apm-host-injection.yml
```diff
@@ -0,0 +1,35 @@
+---
```
bkabrda: Only the Agent 7 repositories contain APM packages. We should check that `datadog_agent_major_version | int > 6` here and fail if it's not.

tasks/pkg-debian/install-apm-host-inject-latest.yml
```diff
@@ -0,0 +1,15 @@
+---
```
bkabrda: All tasks in this file should have `when: not ansible_check_mode`

tasks/pkg-redhat/install-apm-host-inject-latest.yml
```diff
@@ -0,0 +1,16 @@
+---
```
bkabrda: All tasks in this file should have `when: not ansible_check_mode`

tasks/apm-host-injection.yml
```diff
@@ -0,0 +1,35 @@
...
+
+# TODO: enable when flag is ready
+#- name: Check if dd-host-install needs to be run
+#  command: dd-host-install --dry-run
+#  register: dd_host_install_cmd_result
+#  ignore_errors: true
+
+- name: Run dd-host-install
+  command: dd-host-install
+  # when: dd_host_install_cmd_result.rc != 0
```
bkabrda: I think we want to restart the Agent at the end of this file, correct? Because at this point we have installed all the packages and potentially changed the config. We also have to figure out proper conditions on when to actually restart, because we should only restart when something relevant has changed on the host (for example, if we needed to rerun the `dd-host-install` script, we probably need to restart; if we needed to update/install a single language tracing library, we probably don't need to restart).

tasks/main.yml
```diff
@@ -69,3 +69,7 @@
 - name: Integrations Tasks
   include_tasks: integration.yml
   when: datadog_integration is defined
+
+- name: APM Host injection tasks
+  include_tasks: apm-host-injection.yml
+  when: datadog_apm_host_injection_enabled
```
bkabrda: This should be `when: datadog_apm_host_injection_enabled or datadog_apm_docker_injection_enabled`, should it not?

bkabrda (changes requested):
Hi :wave: thanks for sending a PR. I think this is a good start, but it will need some more work. I left comments inline addressing what needs to be done.

tasks/apm-host-injection.yml
```diff
@@ -0,0 +1,35 @@
...
+  when: ansible_facts.os_family in ["RedHat", "Rocky", "AlmaLinux"]
+
+# TODO: enable when flag is ready
+#- name: Check if dd-host-install needs to be run
+#  command: dd-host-install --dry-run
+#  register: dd_host_install_cmd_result
+#  ignore_errors: true
+
+- name: Run dd-host-install
+  command: dd-host-install
```
bkabrda: One more note here - we shouldn't let the dd-host-install command change the Agent configuration file, because that is controlled by Ansible. We should add the necessary Agent configuration through Ansible and only configure the rest of the system with the script.

tasks/apm-host-injection.yml
rarguelloF: Official [doc](https://docs.datadoghq.com/tracing/trace_collection/library_injection_local/?tab=agentonhostappincontainers#install-the-preload-library) says it's necessary, but will confirm if it actually is 👍

tasks/apm-host-injection.yml
rarguelloF: For the 1st point, there's already a conditional to include this whole file using that flag: 
```yaml
- name: APM Host injection tasks
  include_tasks: apm-host-injection.yml
  when: datadog_apm_host_injection_enabled
```

For the 2nd, I didn't implement docker injection yet but I will start now!

For the 3 and 4th point -> 👍

bkabrda: Ad first point: what I mean is that when we have both host injection and docker injection options implemented, this shouldn't run when only docker injection is specified, right? Does that make sense?

tasks/apm-host-injection.yml
bkabrda: Another thing to check that I just realized is that we want to make sure we're running on a supported platform, which right now is Debian/Ubuntu and RH-based distros (so not SUSE or Windows).

All these checks should be probably done early in the role execution to ensure we fail fast instead of failing the role at the end and leaving a half-configured host.

tasks/apm-host-injection.yml
rarguelloF: Confirmed it's not necessary for the agent to be running

tasks/apm-host-injection.yml
rarguelloF: When docker injection is enabled, there's a separate command called `dd-host-container-install`. I've updated the PR with docker injection implemented, let me know if it's clear now!

tasks/apm-host-injection.yml
rarguelloF: Updated PR!

tasks/main.yml
rarguelloF: correct!

tasks/apm-host-injection.yml
rarguelloF: updated!

tasks/pkg-debian/install-apm-inject-latest.yml
```diff
@@ -0,0 +1,15 @@
+---
+- name: Install the APM inject library
```
bkabrda: In order to make the installation more performant, can we use just one task to install multiple packages? Doing one APT invocation saves a lot of time in the run when compared to multiple ones (same applies for the RPM installation).

The `name` argument accepts a list of packages, a loop shouldn't be necessary.

defaults/main.yml
```diff
@@ -60,6 +60,9 @@ datadog_agent_version: ""
 # Default Package name for APT and RPM installs - can override in playbook for IOT Agent
 datadog_agent_flavor: "datadog-agent"
 
+# Default Package name for APM Library Injection APT and RPM installs - can override in playbook for IOT Agent
```
bkabrda: Can you explain what you mean by "can override in playbook for IOT Agent"?

tasks/pkg-redhat/install-apm-inject-latest.yml
```diff
@@ -0,0 +1,15 @@
+---
+- name: Install the APM inject library
+  yum:
+    name: datadog-apm-inject
+    state: latest # noqa 403
+    update_cache: yes
+  become: true
```
bkabrda: I don't think we should use `become: true`. We expect these things to be configured at playbook/host level.

tasks/pkg-redhat/install-apm-inject-latest.yml
```diff
@@ -0,0 +1,15 @@
+---
+- name: Install the APM inject library
```
bkabrda: As noted in the APT installation file, let's try to collapse the installation tasks into a single one to save execution time.

tasks/pkg-redhat/install-apm-inject-latest.yml
```diff
@@ -0,0 +1,15 @@
+---
```
bkabrda, line 1: Could we mimic what we have in the `tasks/pkg-redhat/install-latest.yml` and have one task for `dnf` and another one for `yum`?

bkabrda (changes requested):
Hi :wave: thanks for the additional work, this looks much better now! I left couple comments to address inline. I would also like to ask you to add the descriptions of all newly added configuration values to the `README.md` file.

tasks/apm-inject-check.yml
```diff
@@ -0,0 +1,48 @@
...
+
+- name: Check if docker daemon config dir exists
+  stat:
+    path: /etc/docker
+  register: docker_daemon_config_dir
+
+- name: Fail if APM Host container injection requirements are met (Docker installed)
+  fail:
+    msg: /etc/docker does not exist. Please ensure docker is installed or disable the datadog_apm_host_docker_injection_enabled flag.
+  when: datadog_apm_host_docker_injection_enabled and not docker_daemon_config_dir.stat.isdir
```
bkabrda: I just got this error while testing (docker not installed):

```
TASK [ansible-datadog : Fail if APM Host container injection requirements are met (Docker installed)] ************
fatal: [localhost]: FAILED! => {"msg": "The conditional check 'datadog_apm_host_docker_injection_enabled and not docker_daemon_config_dir.stat.isdir' failed. The error was: error while evaluating conditional (datadog_apm_host_docker_injection_enabled and not docker_daemon_config_dir.stat.isdir): 'dict object' has no attribute 'isdir'\n\nThe error appears to be in '/home/slavek.kabrda/programming/ansible-datadog/tasks/apm-inject-check.yml': line 17, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Fail if APM Host container injection requirements are met (Docker installed)\n  ^ here\n"}
```

According to [stat docs](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/stat_module.html), you should do something like

```
when: datadog_apm_host_docker_injection_enabled and (docker_daemon_config_dir.stat.isdir is not defined or not docker_daemon_config_dir.stat.isdir)
```

tasks/apm-inject-install.yml
```diff
@@ -0,0 +1,40 @@
...
+  command: dd-host-container-install --no-config-change --no-agent-restart
+  notify: restart datadog-agent
+  when: not ansible_check_mode and datadog_apm_host_docker_injection_enabled
+  # when: not ansible_check_mode and dd_host_install_cmd_result.rc != 0
+
+- name: Run APM docker injection setup script
+  # dd-container-install will also reload docker as it changes /etc/docker/daemon.json
+  command: dd-container-install --no-config-change --no-agent-restart
+  notify: restart datadog-agent
+  when: not ansible_check_mode and datadog_apm_docker_injection_enabled
```
bkabrda: The variable `datadog_apm_docker_injection_enabled` doesn't seem to be defined anywhere

defaults/main.yml
```diff
@@ -163,11 +166,16 @@ datadog_windows_config_root: "{{ ansible_facts.env['ProgramData'] }}\\Datadog"
...
+# Configure APM host injection.
+datadog_apm_instrumentation_enabled: ""
+# List of apm libraries to install if host or docker injection is enabled.
+# You can see the available values in our official docs: https://docs.datadoghq.com/tracing/trace_collection/library_injection_local
+datadog_apm_instrumentation_languages: ["all"]
 #
 # Internal variables
 # The following variables are for internal use only, do not modify them.
 #
+datadog_apm_instrumentation_rc: 1
```
bkabrda: Internal variables shouldn't be defined here (some were and still are historically and unfortunately that leads to various user issues). Please define it using `set_fact` task, probably in `tasks/apm-inject-check.yml`

tasks/apm-inject-check.yml
```diff
@@ -0,0 +1,48 @@
+---
```
bkabrda, line 1: I would suggest checking here that `datadog_apm_instrumentation_enabled` is one of `all`, `host`, `docker` and failing early if it is not.

tasks/apm-inject-install.yml
```diff
@@ -0,0 +1,39 @@
+---
+- name: Include APM host injection Debian install tasks
+  include_tasks: pkg-debian/install-apm-inject-latest.yml
+  when: not ansible_check_mode and ansible_facts.os_family == "Debian" and datadog_apm_instrumentation_enabled != ""
```
bkabrda: This whole file is only included when `datadog_apm_instrumentation_enabled != ""`, so let's not check this specific condition again throughout this file.

tasks/apm-inject-install.yml
```diff
@@ -0,0 +1,39 @@
...
+  include_tasks: pkg-redhat/install-apm-inject-latest.yml
+  when: not ansible_check_mode and ansible_facts.os_family in ["RedHat", "Rocky", "AlmaLinux"] and datadog_apm_instrumentation_enabled != ""
+
+#- name: Check if dd-host-install needs to run
+#  command: dd-host-install --dry-run
+#  register: datadog_apm_instrumentation_rc
+#  ignore_errors: true
+
+- name: Run APM host injection setup script
+  command: dd-host-install --no-config-change --no-agent-restart
```
bkabrda, line 17: Should this have `--no-agent-config-change` instead of `--no-config-change`?

tasks/apm-inject-install.yml
```diff
@@ -0,0 +1,39 @@
...
+  notify: restart datadog-agent
+  when: not ansible_check_mode and datadog_apm_instrumentation_rc == 1 and datadog_apm_instrumentation_enabled == "host"
```
bkabrda: I'm not sure whether I understand the logic behind the `datadog_apm_instrumentation_rc` variable. Could you please explain it? When would we set it to a different value and how would this all work then?

tasks/apm-inject-install.yml
```diff
@@ -0,0 +1,39 @@
...
+#- name: Check if dd-host-container-install needs to run
+#  command: dd-host-install --dry-run && dd-container-install --dry-run
+#  register: datadog_apm_instrumentation_rc
+#  ignore_errors: true
+
+- name: Run APM host-docker injection (Docker) setup script
+  # dd-host-container-install will also reload docker as it changes /etc/docker/daemon.json
+  command: dd-host-install --no-agent-config-change --no-agent-restart && dd-container-install --no-config-change --no-agent-restart
+  notify: restart datadog-agent
+  when: not ansible_check_mode and datadog_apm_instrumentation_rc == 1 and datadog_apm_instrumentation_enabled == "all"
```
bkabrda: Instead of having this task, why don't we run the other two with e.g. `when: datadog_apm_instrumentation_enabled in ["all", "host"]` and `when: datadog_apm_instrumentation_enabled in ["all", "docker"]`?

tasks/apm-inject-check.yml
rarguelloF: added in line 12

tasks/apm-inject-install.yml
rarguelloF: the flag for the `dd-host-install` command is called `--no-config-change`

tasks/pkg-redhat/install-apm-inject-latest.yml
rarguelloF: changed!

defaults/main.yml
rarguelloF: I will leave this one to @jonbodner

defaults/main.yml
rarguelloF: removed!

tasks/apm-inject-install.yml
rarguelloF: removed check

tasks/apm-inject-install.yml
rarguelloF: I've changed this to use `register` in the `--dry-run` command execution, please take a look!

tasks/apm-inject-install.yml
rarguelloF: sounds good! changed 👍

defaults/main.yml
```diff
@@ -166,6 +169,21 @@ datadog_windows_config_root: "{{ ansible_facts.env['ProgramData'] }}\\Datadog"
 # arguments to supply to the windows installer.
 win_install_args: " "
 
+# Configure APM host injection. Possible values are: "all", "host" or "docker".
+datadog_apm_instrumentation_enabled: ""
+
+# List of apm libraries to install if host or docker injection is enabled.
```
hestonhoffman:
```suggestion
# List of APM libraries to install if host or Docker injection is enabled.
```

hestonhoffman (approved):
Small edit

README.md
```diff
@@ -34,43 +34,47 @@ To deploy the Datadog Agent on hosts, add the Datadog role and your API key to y
...
+| `datadog_macos_download_url`                | Override the URL to download the DMG installer from (macOS only).                                                                                                                                                                                                                                                                                                                                                                                 |
+| `datadog_inject_apm_flavor`                 | Default Package name for APM Library Injection APT and RPM installs - can override in playbook for IOT Agent.                                                                                                                                                                                                                                                                                                                                     |
```
bkabrda: I'm leaning towards not documenting this variable. Even though it's technically overridable which is probably a good idea for solving some theoretical corner cases in the future, I think documenting it right now is more confusing than not. Let's remove this line.

README.md
```diff
@@ -34,43 +34,47 @@ To deploy the Datadog Agent on hosts, add the Datadog role and your API key to y
...
+| `datadog_apm_instrumentation_enabled`       | Configure APM host injection. Possible values are: `all`, `host` or `docker`.                                                                                                                                                                                                                                                                                                                                                                     |
```
bkabrda: Let's provide a brief explanation for the three options.

tasks/pkg-debian/install-apm-inject-latest.yml
```diff
@@ -0,0 +1,12 @@
+---
+- name: Set APM injection install packages
+  set_fact:
+    _dd_apm_install_pkgs: "{{ (_dd_apm_install_pkgs | default([])) + ['datadog-apm-library-' + item] }}"
```
bkabrda: I think just to make sure this works fine with all Jinja versions, we should use `default([], true)` at all instances of usage of this variable (see https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_filters.html#providing-default-values).

tasks/apm-inject-check.yml
bkabrda: This still needs to be fixed.

bkabrda (changes requested):
Hi :wave: this now looks almost ready, but I still found some problems that should be addressed before I can approve.

bkabrda (approved):
LGTM now, great job!

maxiangelo:
Hi didnt you forgot the [Agent on host, app in containers](https://docs.datadoghq.com/tracing/trace_collection/library_injection_local/?tab=agentonhostappincontainers#) possibillity of injecting APM?

PR #486, opened 2023-06-02
Modify integration updates task to prevent always changed status

KevinFairise2:
Tasks responsible for installing or removing Datadog integrations always reported changed even if the integration was already up to date or already removed. 
Now the task will not have a "changed" status if the integration is already installed with the correct version or if it is already removed.

Tested on Ubuntu and Windows, works as expected

### DIFF
--- a/tasks/integration/_linux-macos-shared.yml
+++ b/tasks/integration/_linux-macos-shared.yml
@@ -1,6 +1,20 @@
+# Check current state of the integrations
+
+- name: Check {{ item.key }} integration state
+  command:
+    argv:
+      - "{{ datadog_agent_binary_path }}"
+      - integration
+      - show
+      - -q
+      - "{{ item.key }}"
+  register: integration_version
+  failed_when: false # This task is supposed to fail when the integration is not installed
+  changed_when: false
+
 # Remove integrations
 
-- name: Removing integrations
+- name: Removing {{ item.key }} integration
   command:
     argv:
       - "{{ datadog_agent_binary_path }}"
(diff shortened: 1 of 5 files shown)


tasks/integration/_linux-macos-shared.yml
```diff
@@ -9,16 +23,18 @@
       - "{{ item.key }}"
   become: yes
   become_user: "{{ integration_command_user }}"
-  loop: "{{ datadog_integration|dict2items }}"
-  when: item.value.action == "remove"
+  when: 
+    - item.value.action == "remove"
+    - integration_version.stdout | length > 0
```
bkabrda: I think it's probably more usual to check the return code (`.rc`) rather than output length.

tasks/integration/linux.yml
```diff
@@ -9,3 +9,4 @@
 
 - name: Include shared integration installation and removal tasks
   include_tasks: integration/_linux-macos-shared.yml
+  loop: "{{ datadog_integration|dict2items }}"
```
bkabrda: What does the output look like when trying to install multiple integrations through `include_tasks` in a loop? Will it log the exact same task name every single time? If so, would it be possible to parametrize the name to make the output more meaningful?

bkabrda (changes requested):
Hi :wave: thanks for submitting the PR. For the most part it looks really good, but I left couple notes inline to address.

tasks/integration/linux.yml
KevinFairise2: Actually, the task names that are displayed are the one defined in the included task. Here the "Include shared integration ..." is displayed once and the three task names defined in `tasks/integration/_linux-macos-shared.yml` are displayed for each integration in the loop.

image

tasks/integration/_linux-macos-shared.yml
KevinFairise2: Yes you're right, thanks! I will change this.

bkabrda (approved):
Nice work, approved :+1:

PR #487, opened 2023-06-02
AP-1597 Upgrade ansible-lint version to 6.17

KevinFairise2:
Upgrade ansible-lint to 6.17 and fixes linting issues

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -1,0 +1,3 @@
+# We ignore the FQCN rule on the role repository since FQCN is not compatible with older version of Ansible
+skip_list:
+    - fqcn
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -1,3 +1,4 @@
+---
 # We ignore the FQCN rule on the role repository since FQCN is not compatible with older version of Ansible
 skip_list:
-    - fqcn
+  - fqcn
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -1,4 +1,4 @@
 ---
-# We ignore the FQCN rule on the role repository since FQCN is not compatible with older version of Ansible
+# We ignore the FQCN rule on the role repository since FQCN is not compatible with older version of Ansible (Ansible =< 2.9)
 skip_list:
   - fqcn
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -122,7 +122,7 @@ jobs:
       - checkout
       # TODO: upgrade ansible-lint to latest (will require updating the image for this task)
       - run: pip install ansible-lint==6.17
-      - run: ansible-lint -v .
+      - run: ansible-lint -v tasks/ handlers/
 
   test_install_downgrade:
     parameters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -121,7 +121,7 @@ jobs:
     steps:
       - checkout
       # TODO: upgrade ansible-lint to latest (will require updating the image for this task)
-      - run: pip install ansible-lint==5.4.0
+      - run: pip install ansible-lint==6.17
       - run: ansible-lint -v .
 
   test_install_downgrade:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -122,7 +122,7 @@ jobs:
       - checkout
       # TODO: upgrade ansible-lint to latest (will require updating the image for this task)
       - run: pip install ansible-lint==6.17
-      - run: ansible-lint -v tasks/ handlers/
+      - run: ansible-lint -v --exclude=.circleci/config.yml .
 
   test_install_downgrade:
     parameters:
--- a/handlers/main-macos.yml
+++ b/handlers/main-macos.yml
@@ -8,20 +8,23 @@
 # to finish and if it's in progress for a longer time, bootstrap fails. We use the old
 # unload/load combo because they actually wait.
 - name: Unload datadog-agent service
-  command: "launchctl unload -wF {{ datadog_macos_system_plist_file_path }}"
-  become: yes
-  check_mode: no
+  command: launchctl unload -wF {{ datadog_macos_system_plist_file_path }}
+  become: true
+  check_mode: false
   when: datadog_enabled and not ansible_check_mode and ansible_facts.os_family == "Darwin"
+  changed_when: true
 
 - name: Load datadog-agent service
-  command: "launchctl load -wF {{ datadog_macos_system_plist_file_path }}"
-  become: yes
-  check_mode: no
+  command: launchctl load -wF {{ datadog_macos_system_plist_file_path }}
+  become: true
+  check_mode: false
   when: datadog_enabled and not ansible_check_mode and ansible_facts.os_family == "Darwin"
+  changed_when: true
 
 - name: Restart datadog-agent service
-  command: "launchctl kickstart -k system/{{ datadog_macos_service_name }}"
-  become: yes
+  command: launchctl kickstart -k system/{{ datadog_macos_service_name }}
+  become: true
   register: command_result
-  check_mode: no
+  check_mode: false
   when: datadog_enabled and not ansible_check_mode and ansible_facts.os_family == "Darwin"
+  changed_when: true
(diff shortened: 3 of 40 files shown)


.ansible-lint
```diff
@@ -0,0 +1,4 @@
+---
+# We ignore the FQCN rule on the role repository since FQCN is not compatible with older version of Ansible
```
bkabrda: It would be nice to mention the precise version here for future reference.

.circleci/config.yml
```diff
@@ -121,8 +121,8 @@ jobs:
     steps:
       - checkout
       # TODO: upgrade ansible-lint to latest (will require updating the image for this task)
-      - run: pip install ansible-lint==5.4.0
-      - run: ansible-lint -v .
+      - run: pip install ansible-lint==6.17
+      - run: ansible-lint -v tasks/ handlers/
```
bkabrda: Why do we run only on tasks and handlers? Why not e.g. defaults? I think we should run on as many files as possible (and probably explicitly exclude what shouldn't be linted, rather than explicitly include what should be linted)

handlers/main-macos.yml
```diff
@@ -8,20 +8,23 @@
 # to finish and if it's in progress for a longer time, bootstrap fails. We use the old
 # unload/load combo because they actually wait.
 - name: Unload datadog-agent service
-  command: "launchctl unload -wF {{ datadog_macos_system_plist_file_path }}"
```
bkabrda, line 11: This one is interesting. Is ansible-lint actually recommending to remove the quotes? Can you share what it complains about when these quotes are here?

handlers/main-macos.yml
```diff
@@ -8,20 +8,23 @@
...
-  become: yes
-  check_mode: no
+  command: launchctl unload -wF {{ datadog_macos_system_plist_file_path }}
+  become: true
+  check_mode: false
   when: datadog_enabled and not ansible_check_mode and ansible_facts.os_family == "Darwin"
+  changed_when: true
```
bkabrda, line 15: Why does ansible-lint want us to use `changed_when: true` here?

tasks/_apt-key-import.yml
```diff
@@ -18,71 +19,75 @@
...
     dest: "{{ tempdir.path }}/single.gpg"
     mode: "0600"
-    remote_src: yes
+    remote_src: true
   when: key_fingerprint == datadog_apt_key_current_name and key_needs_import
   changed_when: false
 
-- name: "Import key {{ key_fingerprint }} to {{ datadog_apt_usr_share_keyring }} keyring"
-  shell: "cat {{ tempdir.path }}/single.gpg | gpg --no-default-keyring --keyring {{ datadog_apt_usr_share_keyring }} --import --batch"  # noqa risky-shell-pipe
+- name: Import key to keyring, key,keyring={{ key_fingerprint, datadog_apt_usr_share_keyring }}
```
bkabrda, line 84: Interesting, is ansible-lint telling us to use this specific format when using variables in task names?

tasks/pkg-macos.yml
```diff
@@ -17,70 +17,76 @@
...
   shell:
-    cmd: '/usr/sbin/installer -pkg "`find "/Volumes/datadog_agent" -name \*.pkg 2>/dev/null`" -target /'
-    chdir: '/'
-  become: yes
+    cmd: /usr/sbin/installer -pkg "`find "/Volumes/datadog_agent" -name \*.pkg 2>/dev/null`" -target /
+    chdir: /
+  become: true
   register: datadog_agent_install
   when: (not datadog_skip_install) and (not ansible_check_mode) and (download_dmg_result.status_code == 200)
+  changed_when: true
```
bkabrda, line 72: Why does ansible-lint want us to use `changed_when: true` on this task and the above ones?

tasks/pkg-windows.yml
```diff
@@ -31,18 +33,20 @@
...
-- include_tasks: pkg-windows-opts.yml
 
-- name: pre-Delete temporary msi
+- name: Include Windows opts tasks
+  include_tasks: pkg-windows-opts.yml
+
+- name: Pre-Delete temporary msi
   win_file:
-    path: '%TEMP%\ddagent.msi'
+    path: "%TEMP%\\ddagent.msi"
```
bkabrda: Are you sure these replacements of double backslash with a single one do what they are supposed to do, e.g. producing a correct effective path `%TEMP%\ddagent.msi`?

bkabrda (commented):
:wave: thanks for your work on this. Overall I think looks good, but there are couple of points I'd like you to take a closer look at - I left comments inline. Specifically I'm wondering about the added `changed_when: true` which will effectively add tasks that will always report as changed (undesirable) and also the Windows tasks where we replaced double backslash with a single one - we should test this and ensure it actually still works.

tasks/_apt-key-import.yml
KevinFairise2: Yes indeed, ansible-lint wants all the variable in the name of the task to be put at the end. Because it is easier to search for the task name when all the variables are at the end.

https://ansible-lint.readthedocs.io/rules/name/

handlers/main-macos.yml
KevinFairise2: This one is a bit strange, actually when the quotes are there ansible-lint does not complain about it. But when using the --write options to automatically fix some linting issues, it automatically removes the quotes. Thus we can decide to keep them and it will not output any error. What do you think about it ?

tasks/pkg-windows.yml
KevinFairise2: These were changed automatically by ansible-lint --write, it appears that reverting these changes has no impact on ansible-lint. I will revert it , because I think it is safer to keep it this way.

.ansible-lint
KevinFairise2: You're right, I just updated it

handlers/main-macos.yml
KevinFairise2: Actually, for tasks where Ansible cannot tell by itself if something changed (basically command and shell tasks), ansible-lint force us to add the changed_when directive, to indicate a condition when the task must be considered as changed. Here I think that each time the `when` condition is met, the task will change something (unload the agent). From what I understand of this task it should either be skipped when the `when` condition is not met or reported as `changed`

handlers/main-macos.yml
bkabrda: I don't have a strong preference, but I'd probably lean towards leaving the quotes there. Making huge PRs smaller makes them easier to review and it will also make navigating git history easier.

handlers/main-macos.yml
bkabrda: Right, it's true that the handlers should always have `changed_when: true`, because they actually will restart the Agent. We should just make sure we're triggering (or not triggering) the handlers in the correct situations.

meta/main.yml
```diff
@@ -24,44 +24,47 @@ galaxy_info:
         - bullseye
     - name: EL
       versions:
-        - 8
-        - 7
-        - 6
-    - name: Amazon Linux 2
+        - '7'
```
bkabrda, line 28: Why did we leave out `6` here?

meta/main.yml
```diff
@@ -24,44 +24,47 @@ galaxy_info:
...
-        - 2016
-        - 2019
-    - name: Darwin
-      versions:
-        - any
+        - '2016'
+        - '2019'
+    # Playbook works on Darwin platform, but we need to comment this out
+    # since schema used by ansible-lint to make validation do not reference Darwin:
+    # https://github.com/ansible/ansible-lint/blob/main/src/ansiblelint/schemas/meta.json
```
bkabrda: Correct, but it does seem to have macos and macosx. These are probably more suitable here? (Not sure what's the difference, but you could probably do the research on that?)

tasks/_apt-key-import.yml
bkabrda: Interesting, thanks for the pointer.

tasks/pkg-macos.yml
bkabrda: Ok, so I reached the conclusion that the `changed_when: true` in this file are actually ok. These tasks will be skipped if `datadog_skip_install` is set to `true`, but otherwise they will actually do things that (possibly) change things on the host (and then they'll trigger the macos handlers to restart the agent, which again will actually execute commands which restart the agent).

bkabrda (approved):
LGTM now, nice job :tada:

PR #490, opened 2023-06-12
Ansible-lint now runs on a collection instead of a role

Pythyu:
The role create a collection at every pipeline to have a matching ansible-lint output with the collection. All variable has been renamed into the ``rolename_varname``  format.

## QA
  - CI :
    - CI now check ansible-lint as in a collection and as a role (in a role we have to exclude a few folder [see](https://github.com/DataDog/ansible-datadog/pull/490/files#diff-c2e20c78d6a1e9721a01e31217f318a569645d1a6f64d80f6425b4d63e36ecffR6) )
    - CI check the installation of the agent via ansible on multiple system 
    - CI **MUST** be green

  - Manual Testing : 
    - From the root of the git repository run ``./ci_test/collection-structure-setup.sh``. You should have no lint issues on the two ansible-lint check (one for the role, one for the collection). It will create a ``ansible_collections/`` folder containing the collection.
    - Go to the newly created ``ansible_collections/datadog/dd`` repository and run ``ansible-galaxy collection build``  it should generate a ``datadog-dd-1.0.0.tar.gz `` file.
    - Run locally at least one manual scenario locally (prefer the ``_full.yml``tests)

## QA Instructions

Follow and check that the previous process is working properly. Be careful that every option has been processed.
Additionally, install the agent on one or more ec2 instance to check the full ansible process. Preferably with a maximum of option in the playbook to ensure a maximum amount of variable are correctly acknowledged.


.circleci/config.yml
```diff
@@ -120,9 +120,12 @@ jobs:
       - image: datadog/docker-library:ansible_debian_2_10
     steps:
       - checkout
-      # TODO: upgrade ansible-lint to latest (will require updating the image for this task)
-      - run: pip install ansible-lint==6.17
-      - run: ansible-lint -v --exclude=.circleci/config.yml .
+      - run: pip install ansible-lint==6.17 galaxy-importer
+      - run: mkdir ansible_collections
+      - run: cd ansible_collections && ansible-galaxy collection init datadog.dd && mkdir -p datadog/dd/roles/agent/
+      - run: for file in $(find . -type f -and \( -path '*defaults*' -or -path '*tasks*' -or -path '*templates*' -or -path '*handlers*' -or -name '.gitkeep' -or -path '*meta*' -or -name 'README.md' \)); do echo "Create ansible_collections/datadog/dd/roles/agent/${file:2}" && mkdir -p $(dirname "ansible_collections/datadog/dd/roles/agent/${file:2}") && cp -r "$file" "ansible_collections/datadog/dd/roles/agent/${file:2}"; done
```
chouetz: This is not easy to read. What about putting this in a dedicated file and source it here? It would maybe help if we have further modifications, wdyt?

defaults/main.yml
```diff
@@ -55,9 +55,9 @@ datadog_additional_groups: {}
 datadog_agent_major_version: ""
 
 # Pin agent to a version. Highly recommended.
-# Defaults to the latest version of the major version chosen in datadog_agent_major_version
-# If both datadog_agent_major_version and datadog_agent_version are set, they must be
-# compatible (ie. the major version in datadog_agent_version must be datadog_agent_major_version)
+# Defaults to the latest version of the major version chosen in agent_datadog_agent_major_version
```
chouetz: There's probably a reason behind, but you have twice `agent` in the same name

manual_tests/test_5_full.yml
```diff
@@ -15,7 +15,7 @@
         env: dev
       trace.concentrator:
         extra_aggregators: version
-    datadog_checks:
+    agent_datadog_checks:
```
chouetz: for english would it be possible to have `datadog_agent` rather than `agent_datadog`?

manual_tests/test_6_full.yml
```diff
@@ -42,7 +42,7 @@
         instances:
           - host: localhost
             port: 22
-            username: root
+            agent_username: root
```
chouetz: for consistency, maybe `agent_datadog` (or `datadog_agent`) here?

manual_tests/test_7_full.yml
```diff
@@ -47,7 +47,7 @@
         instances:
           - host: localhost
             port: 22
-            username: root
+            agent_username: root
```
chouetz: same as previous, for consistency, maybe `agent_datadog` (or `datadog_agent`) here?

tasks/_agent-linux-macos-shared.yml
```diff
@@ -1,99 +1,99 @@
 ---
 - name: Create Datadog agent config directory
   file:
-    dest: "{{ _dd_config_dir }}"
+    dest: "{{ agent__dd_config_dir }}"
```
chouetz: are the two underscores expected? For consistency we can maybe extend dd to datadog?

tasks/_agent-linux-macos-shared.yml
```diff
@@ -1,99 +1,99 @@
...
     state: directory
     mode: "0755"
-    owner: "{{ _dd_user }}"
-    group: "{{ _dd_group }}"
+    owner: "{{ agent__dd_user }}"
```
chouetz: same as previous, maybe `_dd_agent_user` or `_agent_datadog_user` or `_datadog_agent_user` :) ?

tasks/_apt-key-import.yml
```diff
@@ -21,75 +21,76 @@
 
 - name: Set local variables for processed key {{ item.key }}
   set_fact:
-    key_fingerprint: "{{ item.key }}"
-    keyring_url: "{{ item.value }}"
+    agent_key_fingerprint: "{{ item.key }}"
```
chouetz, line 24: `agent` or `agent_datadog` (or `datadog_agent`)?

tasks/agent-linux.yml
```diff
@@ -25,10 +27,10 @@
 - name: Include configuration setup tasks
   include_tasks: _agent-linux-macos-shared.yml
   vars:
-    _dd_config_dir: /etc/datadog-agent
-    _dd_user: "{{ datadog_user }}"
-    _dd_group: "{{ datadog_group }}"
-    _dd_notify_agent: restart datadog-agent
+    agent__dd_config_dir: /etc/datadog-agent
```
chouetz: underscores here again?

defaults/main.yml
Pythyu: All variable must be prefixed with ``agent_`` because of the ``ansible-lint`` expected role variable format in collections. Since a large number of variable were replaced, It was done automatically hence why it replaced comments variable as well.

manual_tests/test_5_full.yml
Pythyu: Can't do see above ^^

manual_tests/test_6_full.yml
Pythyu: same as above, and we want to keep the ``username`` part of it to remain

manual_tests/test_7_full.yml
Pythyu: same as above

tasks/_agent-linux-macos-shared.yml
Pythyu: same as above but double underscore is more of a side-effect of the automated replace, I can fix it into single underscore

tasks/_agent-linux-macos-shared.yml
Pythyu: same as above

tasks/_apt-key-import.yml
Pythyu: I want to keep ``key_fingerprint`` so ``agent_key_fingerprint`` since ``rolename_variablename`` format

tasks/agent-linux.yml
Pythyu: same as above could fix it

.circleci/config.yml
Pythyu: I can give it it's dedicated file sure looks better this way for sure

ci_test/downgrade_to_5.yaml
```diff
@@ -17,7 +17,7 @@
         env: dev
       trace.concentrator:
         extra_aggregators: version
-    datadog_checks:
+    agent_datadog_checks:
```
KevinFairise2: I think this variable name should not change, since this is a public-facing variable

manual_tests/test_7_full.yml
KevinFairise2: Same as above for this one ?

ci_test/downgrade_to_5.yaml
bkabrda: Why are we renaming these variables in `ci_test` and `manual_tests` directories? This should still be `datadog_checks`, because we handle it in `tasks/variable-setup.yml`, correct?

.circleci/config.yml
bkabrda: I would vote for a dedicated file because we want to be able to easily run this test locally. Right now I would probably have to manually repeat the steps done here. Or is there a better way right now?

defaults/main.yml
bkabrda: We can't replace the comment here, because it makes it invalid. This does need to reference `datadog_agent_major_version`.

ci_test/downgrade_to_5.yaml
Pythyu: In [there](https://github.com/DataDog/ansible-datadog/pull/490/files#diff-62f3d41880f6ed7a6e626fa7224fb9bcf32911acccd8d558cd4297a8624c9c96) it's defined so it's normal

manual_tests/test_7_full.yml
Pythyu: I forgot this one will fix it 👍

tasks/pkg-windows-opts.yml
```diff
@@ -22,67 +22,67 @@
...
   when: datadog_windows_config_files_dir | length > 0
 
 # if the current installation was set to a non-standard config root, and that config root is not
 # presented here, then update accordingly, so that any config file modifications will be made
 # in the right place
 - name: Set config root for config Files
   set_fact:
-    datadog_windows_config_root: "{{ datadog_windows_config_files_dir }}"
-  when: ((datadog_windows_config_files_dir | length > 0) and (not config_root_from_registry.exists))
+    agent_datadog_windows_config_root: "{{ datadog_windows_config_files_dir }}"
```
bkabrda, line 76: There's a default `datadog_windows_config_root` in `defaults/main.yml`, which right now will never get used. We have to initialize the `agent_datadog_windows_config_root` fact to its value before this task.

tasks/agent5-linux.yml
```diff
@@ -45,24 +45,24 @@
     patterns:
       - "*.yaml.default"
     file_type: file
-  register: datadog_conf_files_default
+  register: agent_agent_datadog_conf_files_default
```
bkabrda: Why the double `agent_agent_`?

tasks/main.yml
```diff
@@ -10,16 +13,16 @@
 - name: Check if OS is supported
   include_tasks: os-check.yml
 
-- name: Resolve datadog_tracked_checks later to defend against variable presidence issues arising from dynamically included null datadog_checks
+- name: Resolve agent_datadog_tracked_checks later to defend against variable presidence issues arising from dynamically included null agent_datadog_checks
```
bkabrda: The `datadog_checks` in this name refers to the configuration variable, not our internal variable. I believe it should remain as is.

tasks/variable-setup.yml
```diff
@@ -0,0 +1,12 @@
+---
+- name: Set datadog_checks variable
+  set_fact:
+    agent_datadog_checks: "{{ datadog_checks | default({}, true) }}"
+
+- name: Set datadog_agent_major_version variable
```
bkabrda: Do we need this task here? We do exactly the same in `set-parse-version.yml` where it is sufficient AFAICS.

tasks/set-parse-version.yml
```diff
@@ -1,17 +1,17 @@
 ---
-- name: Convert datadog_agent_major_version to string
+- name: Convert agent_datadog_agent_major_version to string
```
bkabrda: We actually convert `datadog_agent_major_version`, not `agent_datadog_agent_major_version`, so let's fix that in the task name.

tasks/pkg-suse.yml
```diff
@@ -69,13 +69,13 @@
     - name: Stat if 20200908 key (Expires 2024) RPM key already exists
       stat:
         path: /tmp/DATADOG_RPM_KEY_20200908.public
-      register: ddnewkey_20200908
+      register: agent_agent_ddnewkey_20200908
```
bkabrda: I don't think we need to use `agent_` twice - probably an oversight here.

tasks/variable-setup.yml
```diff
@@ -0,0 +1,12 @@
...
+  set_fact:
+    agent_datadog_agent_major_version: "{{ datadog_agent_major_version | default('', true) | string }}"
+
+- name: Set datadog_zypper_repo variable
```
bkabrda: Why do we need this one? I don't see it used anywhere.

tasks/variable-setup.yml
```diff
@@ -0,0 +1,12 @@
+---
+- name: Set datadog_checks variable
+  set_fact:
+    agent_datadog_checks: "{{ datadog_checks | default({}, true) }}"
```
bkabrda: We do basically the same in `sanitize-checks.yml`, don't we? I think we can just do it there and not here.

tasks/pkg-windows-opts.yml
```diff
@@ -1,19 +1,19 @@
 ---
 - name: Set DD Username Arg
   set_fact:
-    win_install_args: "{{ win_install_args }} DDAGENTUSER_NAME={{ datadog_windows_ddagentuser_name }}"
+    agent_win_install_args: "{{ agent_win_install_args }} DDAGENTUSER_NAME={{ datadog_windows_ddagentuser_name }}"
```
bkabrda: This variable is defined in `defaults/main.yml`, but I don't see it respected here. We'll have to first set `agent_win_install_args` to value of `win_install_args`.

tasks/sanitize-checks.yml
```diff
@@ -1,13 +1,13 @@
 ---
-- name: Defend against defined but null datadog_checks variable
+- name: Defend against defined but null agent_datadog_checks variable
```
bkabrda: If we leave the definition from `variable-setup.yml` in favor of this task, let's call it `datadog_checks`, not `agent_datadog_checks`.

bkabrda (changes requested):

.circleci/config.yml
```diff
@@ -120,9 +120,9 @@ jobs:
       - image: datadog/docker-library:ansible_debian_2_10
     steps:
       - checkout
-      # TODO: upgrade ansible-lint to latest (will require updating the image for this task)
-      - run: pip install ansible-lint==6.17
-      - run: ansible-lint -v --exclude=.circleci/config.yml .
+      - run: pip install ansible-lint==6.17 galaxy-importer
+      - run: .circleci/collection-structure-setup.sh
+      - run: cd ansible_collections/datadog/dd && ansible-lint -v --profile=production --exclude=galaxy.yml --exclude=meta/
```
bkabrda: Could we make all of this part of the script? What I would like to do is just run e.g. `ci-test/test-collection-lint.sh` and get the same result that CI has. We should put the script somewhere else than the `.circleci` repository, because that makes it seem like it's only for CI usage, but it's actually desirable that this file is used by contributors to check their PRs. I propose to move it to `ci-test` and also mention it in `CONTRIBUTING.md`. Does that make sense?

tasks/pkg-windows-opts.yml
bkabrda: @Pythyu you marked this as resolved, but I don't think it actually is... where have you fixed this?

tasks/main.yml
bkabrda: Again, this is marked as resolved but I don't think it is.

tasks/set-parse-version.yml
bkabrda: Again, I don't think this is actually resolved.

bkabrda (changes requested):
Thanks for doing the updates - I have couple more notes on things that I think we can still improve/fix.

Pythyu:
Ah mb on the solved one thought it was solved. fixing that rn

.circleci/config.yml
Pythyu: Good idea 👍

bkabrda (approved):

KSerrania (commented):
This is a large change. Could you add in the PR description:
- the tests you made (both automatic and manual) to verify that the behavior of the role didn't change,
- a QA plan of what we should test before releasing this change.

PR #494, opened 2023-06-28
Fix a few forgotten variable that hasn't been renamed

Pythyu:
Circle CI only test major version currently that's the reason why the CI didn't trigger on this issue on the last PR about renaming (agent_datadog_minor undefined)


bkabrda (approved):

chouetz (approved):
ok

PR #496, opened 2023-06-29
Improve conditions for APM scripts dry-run flags

bkabrda:
We only need to run them when the specific instrumentation type (host/docker) is enabled because otherwise they'll be always run even if they're not necessary.


rarguelloF (approved):
Nice improvement, thanks!!

Pythyu (approved):
LGTM !

PR #498, opened 2023-07-06
Use agent_datadog_agent_windows_version instead of datadog_agent_windows_version

KSerrania:
### What does this PR do?

Updates the `win_agent_latest.yml` / `win_agent_version.yml` inclusion tasks to use `agent_datadog_agent_windows_version`.

### Motivation

`datadog_agent_windows_version` doesn't exist anymore, the latest version of the Agent was always installed.


Pythyu (approved):
👍

PR #499, opened 2023-07-07
Use agent_datadog_agent_macos_version instead of datadog_agent_macos_version

Pythyu:
Replace the ``datadog_agent_macos_version`` variable with ``agent_datadog_agent_macos_version`` in ``pkg-macos.yml``. 
Since ``datadog_agent_macos_version``  doesn't exist anymore the latest version of the agent was always installed.


KevinFairise2 (approved):
LGTM

PR #506, opened 2023-07-20
Enable Empty-string-compare rule

KevinFairise2:
Enable the empty-string-compare rule which is disabled by default on recent version of ansible-lint. 
To comply with ansible galaxy-importer linter

Removed `--profile production` from the ansible-lint command, otherwile it does not include opt-int rules included. Should not be a problem since by default it will run on all the profiles.


amenasria (approved):
LGTM !

PR #509, opened 2023-07-24
[Windows] Force reinstall if configuration changed [AP-1946]

chouquette:
This PR adds support for optional features exposed by the MSI installer.
It will probe the available features and will trigger a reinstall if an optional features must be enabled/disabled (the installer doesn't support in place modification of the installed features)

While this is of limited interest given that the installer always installs all features since 7.46, this paves the way for future optional features that we might not want to systematically install

### DIFF
--- a/tasks/pkg-windows-opts.yml
+++ b/tasks/pkg-windows-opts.yml
@@ -110,3 +110,16 @@
   set_fact:
     agent_win_install_args: "{{ agent_win_install_args }} ADDLOCAL=MainApplication,NPM"
   when: agent_datadog_sysprobe_enabled
+
+# Check for potential config changes that require a reinstall:
+- name: Check NPM config change
+  set_fact:
+    agent_datadog_skip_install: false
+    agent_datadog_force_reinstall: true
+  # Starting from agent 7.45 the NPM driver is installed by default and gets
+  # reported as such regardless of the provided parameters. Since we don't
+  # want to force the install for nothing in those version, we skip the step
+  # when the agent is 7.45+
+  when: >-
+    (agent_datadog_skip_install and (agent_datadog_major | int <= 7) and (agent_datadog_minor | int < 45)) and
+    (("NPM" in datadog_windows_agent_active_config.stdout_lines) != agent_datadog_sysprobe_enabled)
--- a/tasks/pkg-windows-opts.yml
+++ b/tasks/pkg-windows-opts.yml
@@ -48,7 +48,7 @@
     name: InstallPath
   register: agent_install_path_from_registry
 
-## validate the config path.  Only necessary if it's set in the registry alread (i.e. upgrade)
+## validate the config path.  Only necessary if it's set in the registry already (i.e. upgrade)
 ## Will fail the install if the caller has set the config root to a non-standard root, and that
 ## root is different than what's already present.
 - name: Validate config path
--- a/tasks/pkg-windows-opts.yml
+++ b/tasks/pkg-windows-opts.yml
@@ -12,6 +12,22 @@
     agent_win_install_args: "{{ win_install_args }} DDAGENTUSER_NAME={{ datadog_windows_ddagentuser_name }}"
   when: datadog_windows_ddagentuser_name | default('', true) | length > 0
 
+- name: Get Windows Agent config
+  win_shell: |
+    $installer = new-object -comobject WindowsInstaller.Installer
+    $agent_item = (Get-ItemProperty "HKLM:\Software\Microsoft\Windows\CurrentVersion\Uninstall\*" -EA 0 | Where { $_.DisplayName -like "Datadog Agent*" }).PSChildName
+    if (!$agent_item) {
+      return
+    }
+    $features = $installer.features($agent_item)
+    ForEach ($f in $features) {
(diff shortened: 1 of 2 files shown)


tasks/pkg-windows.yml
```diff
@@ -4,6 +4,31 @@
...
+  win_reg_stat:
+    path: HKLM:\SOFTWARE\Datadog\Datadog Agent
+    name: InstallArgs
+  register: agent_config_install_args
+
+- name: Force install if config changed
+  set_fact:
+    agent_datadog_skip_install: >-
+      {{ agent_config_install_args.exists and
+      agent_config_install_args.value != agent_win_install_args }}
```
KevinFairise2: Shouldn't we skip the install when the provided args match the previous one stored in the registry? It looks like `agent_config_install_args.value != agent_win_install_args` will set `agent_datadog_skip_install` to true when they are not the same.

tasks/pkg-windows.yml
```diff
@@ -67,23 +84,24 @@
...
-    arguments: "{{ agent_win_install_args }}"
+    arguments: >-
+      {{ agent_win_install_args }}
+      {% if datadog_windows_ddagentuser_password | default('', true) | length > 0 %}
+      DDAGENTUSER_PASSWORD={{ datadog_windows_ddagentuser_password }}
+      {% endif %}"
   register: agent_datadog_agent_install
   when: (not agent_datadog_skip_install) and (not ansible_check_mode)
 
+- name: Store successful instal args
```
KevinFairise2:
```suggestion
- name: Store successful install args
```

tasks/pkg-windows.yml
chouquette: Hm, indeed you are correct the logic is inverted, which means something else is wrong since the install was still forced during my tests

tasks/pkg-windows.yml
```diff
@@ -4,6 +4,31 @@
...
+  debug:
+    msg: "{{ agent_win_install_args }}{% if datadog_windows_ddagentuser_password | default('', true) | length > 0 %} DDAGENTUSER_PASSWORD=<REDACTED>{% endif %}"
+
+- name: Collect previous agent install arguments
+  win_reg_stat:
+    path: HKLM:\SOFTWARE\Datadog\Datadog Agent
+    name: InstallArgs
+  register: agent_config_install_args
+
+- name: Force install if config changed
```
KevinFairise2, line 25: I think you would also need to add previous value of agent_datadog_skip_install in the condition:
```
{{ agent_datadog_skip_install and agent_config_install_args.exists and
      agent_config_install_args.value != agent_win_install_args }}
```
Otherwise we could skip the install when we want to install another version of the agent with the same install args

tasks/pkg-windows.yml
chouquette: I found the issue, it was somewhere between the chair and the keyboard (I forgot to pin the version so I was checking with a `agent_datadog_skip_install` set to false, meaning I didn't test anything valuable)

tasks/pkg-windows-opts.yml
```diff
@@ -12,6 +16,23 @@
     agent_win_install_args: "{{ win_install_args }} DDAGENTUSER_NAME={{ datadog_windows_ddagentuser_name }}"
   when: datadog_windows_ddagentuser_name | default('', true) | length > 0
 
+- name: Get Windows Agent config
```
julien-lebot:
```suggestion
- name: Get Windows Agent features
```
The term config for the Agent normally refers to the configuration files, I think this could cause confusion here since this code extracts the installer features.

tasks/pkg-windows-opts.yml
```diff
@@ -90,3 +111,16 @@
   set_fact:
     agent_win_install_args: "{{ agent_win_install_args }} ADDLOCAL=MainApplication,NPM"
   when: agent_datadog_sysprobe_enabled
+
+# Check for potential config changes that require a reinstall:
+- name: Check NPM config change
```
julien-lebot:
```suggestion
- name: Check NPM feature change
```

tasks/pkg-windows-opts.yml
```diff
@@ -90,3 +111,16 @@
...
+  set_fact:
+    agent_datadog_skip_install: false
+    agent_datadog_force_reinstall: true
+  # Starting from agent 7.45 the NPM driver is installed by default and gets
+  # reported as such regardless of the provided parameters. Since we don't
+  # want to force the install for nothing in those version, we skip the step
+  # when the agent is 7.45+
+  when: >-
```
julien-lebot, line 126: 👍 I wonder if we should also print a message to inform the user of this change in behavior ?

tasks/pkg-windows-opts.yml
chouquette: renamed

tasks/pkg-windows-opts.yml
chouquette: renamed

tasks/pkg-windows-opts.yml
chouquette: Added a debug message to let the users know

chouquette:
Thanks for the review! I pushed an additional commit per your suggestions (and rebased on top of main)

julien-lebot (approved):

KevinFairise2 (approved):
LGTM!

PR #512, opened 2023-08-23
Fail early when the API key is missing [AP-2380]

chouquette:


spencergilbert (commented):
Does this need a CHANGELOG addition?

chouquette:
> Does this need a CHANGELOG addition?

I agree, but I *think* we update the changelog when releasing, for this repo at least. 

I suppose we don't really know what the next version will be at this point which is why we wait?

spencergilbert (approved):
Gotcha, I'm used to updating a changelog with "unreleased" as the changes come in. Whatever the standard is works 👍

PR #523, opened 2023-10-30
correct syntax when checking for config changes

TomFromTA:
having both `debug` and `set_fact` in a single task is invalid. This PR places the message details from the `debug` task into the `name`.


chouquette (approved):
Thanks a lot for fixing this!

PR #528, opened 2023-11-13
Change mode to string parameters

### LINKED ISSUE #527: Mode 600 will set incorrect permissions on files.
Using non string 600 is a booby trap. Read the docs here.
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/get_url_module.html#parameter-mode
Newly installed hosts without an rpm key bombs hard on this (RHEL9)

https://github.com/DataDog/ansible-datadog/blob/f53b1ab1d63c99410b832f22208470d23b30d71a/tasks/pkg-redhat.yml#L30-L35

Here is an example playbook:
```yaml
- name: Mode test
  hosts: localhost
  tasks:
    - name: Create file
      file:
        path: /tmp/mode_test
        mode: 600
        state: touch
```
Result:
```
$ ls -l /tmp/mode_test
---x-wx--T 1 user user 0 Nov 13 11:10 /tmp/mode_test
```

Here is a correct playbook:
```yaml
- name: Mode test
  hosts: localhost
  tasks:
    - name: Create file
      file:
        path: /tmp/mode_test
        mode: '600'
        state: touch
```
Result:
```
$ ls -l /tmp/mode_test
-rw------- 1 user user 0 Nov 13 11:11 /tmp/mode_test
```

janorn:
Update playbook to use string type on mode parameter.

Fixes #527

### DIFF
--- a/tasks/_apt-key-import.yml
+++ b/tasks/_apt-key-import.yml
@@ -59,7 +59,7 @@
     url: "{{ agent_keyring_url }}"
     dest: "{{ agent_tempdir.path }}/{{ agent_key_fingerprint }}"
     force: true
-    mode: 600
+    mode: '600'
   when: agent_key_needs_import
   changed_when: false
 
--- a/tasks/pkg-redhat.yml
+++ b/tasks/pkg-redhat.yml
@@ -31,7 +31,7 @@
       get_url:
         url: "{{ datadog_yum_gpgkey_current }}"
         dest: /tmp/DATADOG_RPM_KEY_CURRENT.public
-        mode: 600
+        mode: '600'
         force: true
 
     - name: Import current RPM key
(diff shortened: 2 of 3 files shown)


janorn:
From Ansible docs:

> For those used to /usr/bin/chmod remember that modes are actually octal numbers. You must give Ansible enough information to parse them correctly. For consistent results, quote octal numbers (for example, '644' or '1777') so Ansible receives a string and can do its own conversion from string into number. Adding a leading zero (for example, 0755) works sometimes, but can fail in loops and some other circumstances.

janorn:
Rebased and squashed commits

janorn:
@chouetz anything else needed for this to be reviewed?

bkabrda (approved):
LGTM, merging. Thanks a lot for the PR!

PR #532, opened 2023-11-20
remove support for datadog-apm-library-all meta package

schneideradam:


defaults/main.yml
```diff
@@ -189,7 +189,7 @@ datadog_apm_instrumentation_enabled: ""
 
 # List of APM libraries to install if host or Docker injection is enabled.
 # You can see the available values in our official docs: https://docs.datadoghq.com/tracing/trace_collection/library_injection_local
-datadog_apm_instrumentation_languages: ["all"]
+datadog_apm_instrumentation_languages: ["java", "node", "dotnet", "python", "ruby"]
```
bkabrda: I believe leaving "all" here would be better, because then we have one less place to modify the list if we're adding a new tracer.

bkabrda: Actually, let's do something different here. Let's create an internal variable in `tasks/apm-inject-check.yml` that will have content `["java", "node", "dotnet", "python", "ruby"]`. When we see `"all"` in this variable, we'll just use the content of that new variable - that way, we'll have all tracers defined in just one place. Does that make sense?

bkabrda (approved):

bkabrda (approved):

schneideradam:
@bkabrda I don't have write access, so i'm unable to merge.

PR #533, opened 2023-11-21
Prepare release 4.21

chouetz:


CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 CHANGELOG
 =========
+# 4.21.0 / 2023-11-22
```
bkabrda: Let's make the changelog entries consistent with the format that we've used so far. Please see the previous releases for the types of changes we use (`[FEATURE]` etc).

bkabrda (approved):

PR #536, opened 2023-12-21
Use Get-ItemProperty to retrieve Windows Agent version

### LINKED ISSUE #526: Improve the version check on Windows and avoid slow idempotency.
https://github.com/DataDog/ansible-datadog/blob/f53b1ab1d63c99410b832f22208470d23b30d71a/tasks/parse-version-windows.yml#L7

Maybe I missed something but I think a command like:
```shell
# ~1-2 seconds
# Adapt the "Select-Object" to your need (in this case, maybe just 'DisplayVersion')
Get-ItemProperty HKLM:\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\*, HKLM:\Software\Microsoft\Windows\CurrentVersion\Uninstall\* | Select-Object DisplayName,DisplayVersion,InstallDate,InstallLocation,PSChildName -unique | Sort-Object DisplayName | select-string "Datadog Agent"

# Output
@{DisplayName=Datadog Agent; DisplayVersion=7.20.2.1; InstallDate=20231108; InstallLocation=;
PSChildName={5A7DD234-696B-4999-852E-4176B1CF4F97}}
```

is far more efficient and complete than the actual implementation:
```shell
# ~1-2 minutes
Get-WmiObject -query "Select Name,IdentifyingNumber,InstallDate,InstallLocation,ProductID,Version FROM Win32_Product where Name like 'Datadog Agent'"

# Output
__GENUS           : 2
__CLASS           : Win32_Product
__SUPERCLASS      :
__DYNASTY         :
__RELPATH         :
__PROPERTY_COUNT  : 6
__DERIVATION      : {}
__SERVER          :
__NAMESPACE       :
__PATH            :
IdentifyingNumber : {5A7DD234-696B-4999-852E-4176B1CF4F97}
InstallDate       : 20231108
InstallLocation   :
Name              : Datadog Agent
ProductID         :
Version           : 7.20.2.1
PSComputerName    :
```

On a general note, consider identifying and optimizing slow Ansible steps in an idempotency context (where the target host is already properly installed/configured) because this role is not efficient in that matter (e.g. when this role is declared as a dependency of other roles often called).

KevinFairise2:
This should close #526. 
It changes the Powershell function used to retrieve the version of the Windows Agent installed, which should speed up the playbook execution and avoid slow idempotency.
Thanks @ldemers777

### DIFF
--- a/tasks/parse-version-windows.yml
+++ b/tasks/parse-version-windows.yml
@@ -3,13 +3,15 @@
 - name: Get Windows Agent version
   win_shell: |
     $product_name = "Datadog Agent"
-    $query = "Select Name,IdentifyingNumber,InstallDate,InstallLocation,ProductID,Version FROM Win32_Product where Name like '$product_name%'"
-    $installs = Get-WmiObject -query $query
+    $version=Get-ItemProperty HKLM:\Software\Microsoft\Windows\CurrentVersion\Uninstall\* |
+    Select-Object DisplayName,DisplayVersion,PSChildName -unique |
+    Where-Object {$_.DisplayName -clike $product_name} |
+    Select-Object DisplayVersion -ExpandProperty DisplayVersion
 
-    if (!$installs -or ($installs.Count -eq 0) -or ($installs.Count -gt 1)) {
+    if (!$version) {
       Write-Host ""
     } else {
-      $ddmaj, $ddmin, $ddpatch, $ddbuild = $installs.Version.split(".")
+      $ddmaj, $ddmin, $ddpatch, $ddbuild = $version.split(".")
       Write-Host "$($ddmaj).$($ddmin).$($ddpatch)"
     }
   register: agent_datadog_version_check_win


tasks/parse-version-windows.yml
```diff
@@ -3,13 +3,15 @@
 - name: Get Windows Agent version
   win_shell: |
     $product_name = "Datadog Agent"
-    $query = "Select Name,IdentifyingNumber,InstallDate,InstallLocation,ProductID,Version FROM Win32_Product where Name like '$product_name%'"
-    $installs = Get-WmiObject -query $query
+    $version=Get-ItemProperty HKLM:\Software\Microsoft\Windows\CurrentVersion\Uninstall\* |
```
chouetz, line 6: compared to the initial proposal the `HKLM:\SOFTWARE\WOW6432Node...` part is missing.
I don't know what it covers on top but it could be interesting to include as well, wdyt?

KevinFairise2: I do not think it is required, I mostly took the code from the second comment, that seems working correctly:
https://github.com/DataDog/ansible-datadog/issues/526#issuecomment-1850422352

chouetz (approved):

tasks/parse-version-windows.yml
ldemers777: FYI: Searching in both (with the one 'Wow6432Node') allows for broader support (namely x86 apps on x64). 

More precisely: "Wow6432 means you’re running a 64-bit version of Windows and allows to manage 32-bit applications that runs on a 64-bit version of Windows. When the 32-bit app performs a registry query on HKEY_LOCAL_MACHINE\SOFTWARE\<company>\<product>, Windows automatically redirects the query to HKEY_LOCAL_MACHINE\SOFTWARE\Wow6432Node\<company>\<product>".

Since I carry such commands in many cases (software), it leaked in my snippet/example. I leave this up to you to determine if that can happen (e.g. installing an x86 version of Datadog Agent on a  64-bit version of Windows).

No matter what, thanks for taking care of this issue. It's very appreciated.

KevinFairise2: That's interesting, I did not know about it, the explanation is appreciated! Since we do not provide a 32bits version of the Agent I guess that we do not need to search in `Wow6432Node`

ldemers777: I guess this will eventually land/release on https://github.com/ansible-collections/Datadog as well? It is where we fetch it.

KevinFairise2: Of course, it will. Unfortunately we will not be able to do the release by the end of the year. But we will do it early in January!

KevinFairise2: Sorry for the delay. The new version of the role and the collection was finally released. 
On Ansible Galaxy: 
 - role: https://galaxy.ansible.com/ui/standalone/namespaces/1584/
 - collection: https://galaxy.ansible.com/ui/repo/published/datadog/dd/

PR #541, opened 2024-01-19
Add version pinning and telemetry for APM tracer libraries

schneideradam:
Telemetry to send analytics to datadog for tracer library injection.
Version pinning for specific tracer libraries.

Reference APMON-356


tasks/telemetry.yml
```diff
@@ -0,0 +1,34 @@
+---
+# TODO: @adam.schneider - Make this generic
```
jonbodner: should probably remove the TODO before release.

tasks/telemetry.yml
```diff
@@ -0,0 +1,34 @@
...
+- name: Build APM library installation payloads for telemetry
+  set_fact:
+    telemetry_payload:
+      "{{ telemetry_payload | default([]) |
+          combine({'specified_' + item.split(':')[0] + '_lib_version':
+            item.split(':')[1] | default('default')}) }}"
+  loop: "{{ datadog_apm_instrumentation_libraries }}"
```
jonbodner: is the item name js or nodejs? because there is one case where the  name of the cur_lib doesn't match:

```
        datadog-apm-library-js)
          library_telemetry_info="${library_telemetry_info},\"specified_nodejs_lib_version\": \"${cur_version}\""
```

tasks/telemetry.yml
jonbodner, line 1: Is there a test on the output to make sure that it's writing valid JSON that matches what's expected?

tasks/telemetry.yml
schneideradam: Fixed. Using a map to get specific tag based on the library specified.

tasks/telemetry.yml
schneideradam: I haven't setup a unit test, but in my manual tests the telemetry request body based on
`datadog_apm_instrumentation_libraries: ["java:1.23.0-1", "python:2.0.9-1", "js", "python:3.11-1", "dotnet", "ruby"]`

was...
```
TASK [../../ansible-datadog : View telemetry request] ****************************************************************************************************
ok: [localhost] => {
    "msg": {
        "api_version": "v2",
        "application": {
            "service_name": "apm-ansible-test"
        },
        "payload": {
            "event_name": "agent.injection.success",
            "tags": {
                "install_time": 1707421273,
                "install_type": "ansible",
                "specified_dotnet_lib_version": "default",
                "specified_java_lib_version": "1.23.0-1",
                "specified_nodejs_lib_version": "default",
                "specified_python_lib_version": "3.11-1",
                "specified_ruby_lib_version": "default"
            }
        },
        "request_type": "apm-onboarding-event"
    }
}
```

jhgilbert (approved):

schneideradam:
@jonbodner Feel free to merge this when you have a final look. Thanks!

jonbodner (approved):
LGTM

README.md
```diff
@@ -94,7 +94,7 @@ The API key is required and its absence causes the role to fail. If you want to
 | `datadog_macos_user`                        | The name of the user to run Agent under. The user has to exist, it won't be created automatically. Defaults to `ansible_user` (macOS only).|
 | `datadog_macos_download_url`                | Override the URL to download the DMG installer from (macOS only).|
 | `datadog_apm_instrumentation_enabled`       | Configure APM instrumentation. Possible values are: <br/> - `host`: Both the Agent and your services are running on a host. <br/> - `docker`: The Agent and your services are running in separate Docker containers on the same host.<br/>- `all`: Supports all the previous scenarios for `host` and `docker` at the same time.|
-| `datadog_apm_instrumentation_languages`     | List of APM libraries to install if `host` or `docker` injection is enabled (defaults to `["all"]`). You can find the available values in [Inject Libraries Locally][24].|
```
bkabrda, line 99: In [a similar PR](https://github.com/DataDog/agent-linux-install-script/commit/e7aa1679fecaaf76d12f9cf0cfbb07bca19065be) for the install_script, the old variable wasn't removed. Let's stay consistent and not remove this one either. I would propose removing it from the README, but still respecting it (and printing a deprecation warning when it's used)

tasks/apm-inject-install.yml
```diff
@@ -1,10 +1,19 @@
 ---
+- name: Remove latest tag and append -1 when version number specified in pinned APM tracer libraries
```
bkabrda: Why do we need the `:latest` related functionality? Do we recommend users to use `:latest` somewhere?

tasks/pkg-debian/install-apm-inject.yml
```diff
@@ -0,0 +1,23 @@
+---
+- name: Set APM injection languages and versions for Debian
```
bkabrda, line 2: You should implement version locking, similar to what we did in the [install_script PR](https://github.com/DataDog/agent-linux-install-script/commit/e7aa1679fecaaf76d12f9cf0cfbb07bca19065be#diff-79ac74a50e14d4fa9c80b1daba8bd5cc182f4b6c2a2b3951c0caa4fbc78f6f73R1030) (you should be able to use the [dpkg_selections](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/dpkg_selections_module.html) module)

tasks/pkg-debian/install-apm-inject.yml
```diff
@@ -0,0 +1,23 @@
...
+  set_fact:
+    agent_dd_apm_install_pkgs: "{{ (agent_dd_apm_install_pkgs |
+      default([], true)) + ['datadog-apm-library-' + (item | regex_replace('[:]', '='))] }}"
+  loop: "{{ pre_processed_datadog_apm_libraries }}"
+
+- name: Install APM Inject Library
```
bkabrda, line 16: It would be ideal to collapse the two `apt` tasks into 1 if possible. Each additional `apt` call means longer runtime.

tasks/pkg-redhat/install-apm-inject.yml
```diff
@@ -0,0 +1,38 @@
+---
+- name: Set APM injection languages and versions for Redhat
```
bkabrda, line 3: Ditto with locking packages (done in install_script [here](https://github.com/DataDog/agent-linux-install-script/commit/e7aa1679fecaaf76d12f9cf0cfbb07bca19065be#diff-79ac74a50e14d4fa9c80b1daba8bd5cc182f4b6c2a2b3951c0caa4fbc78f6f73R889))

tasks/pkg-debian/install-apm-inject.yml
```diff
@@ -0,0 +1,23 @@
...
+  apt:
+    name: 'datadog-apm-inject'
+    state: latest # noqa package-latest
+    update_cache: true
+    cache_valid_time: "{{ datadog_apt_cache_valid_time }}"
+
+- name: Install APM inject libraries
+  apt:
```
bkabrda, line 25: The `apt` tasks in this file should have `when: not ansible_check_mode`

tasks/telemetry.yml
```diff
@@ -0,0 +1,42 @@
...
+      java: specified_java_lib_version
+      js: specified_nodejs_lib_version
+      dotnet: specified_dotnet_lib_version
+      python: specified_python_lib_version
+      ruby: specified_ruby_lib_version
+
+- name: Build APM library installation payloads for telemetry
+  set_fact:
+    telemetry_payload:
+      "{{ telemetry_payload | default([]) |
```
bkabrda: Should the `default` here provide a dict instead of list? `combine` is to be applied to dicts AFAICS.

tasks/telemetry.yml
```diff
@@ -0,0 +1,42 @@
...
+      api_version: 'v2'
+      application:
+        service_name: "{{ ansible_hostname }}"
+      payload:
+        event_name: 'agent.injection.success'
+        tags: "{{ {'install_time': ansible_date_time.epoch | int,
+                   'install_type': 'ansible'} |
+                   combine(telemetry_payload) }}"
+
+- name: Send installation telemetry to Datadog
```
bkabrda, line 32: Let's add `when: not ansible_check_mode` for this task.

tasks/telemetry.yml
```diff
@@ -0,0 +1,42 @@
...
+  uri:
+    url: "{{ datadog_apm_telemetry_endpoint }}"
+    method: POST
+    return_content: true
+    body_format: json
+    body: "{{ telemetry_request }}"
+    headers:
+      DD-API-KEY: "{{ datadog_api_key }}"
+    status_code: 202
+  register: telemetry_response
```
bkabrda: I don't think we need to register the variable if we're not using it anywhere.

tasks/telemetry.yml
```diff
@@ -0,0 +1,42 @@
...
+      application:
+        service_name: "{{ ansible_hostname }}"
+      payload:
+        event_name: 'agent.injection.success'
+        tags: "{{ {'install_time': ansible_date_time.epoch | int,
+                   'install_type': 'ansible'} |
+                   combine(telemetry_payload) }}"
+
+- name: Send installation telemetry to Datadog
+  uri:
```
bkabrda, line 33: Should we mark this task with `failed_when: false`? I think we don't want Ansible to fail if for whatever reason we fail to send telemetry. It should also be marked with `changed_when: false`

bkabrda (changes requested):
Hi :wave: I think this looks pretty good, but I have some comments/suggestions for improvements that I left inline. Can you please get these addressed?

tasks/apm-inject-install.yml
schneideradam: This was a request from the product team to catch if a user entered "latest" as the version.

tasks/pkg-debian/install-apm-inject.yml
schneideradam: Yes, I agree. However due to the nature of having a mixture of both versioned, and "latest" packages, it makes more sense to do it this way since ansible needs the `state: latest` for latest, and `state: present` for versioned packages. I realize this is less than optimal, but really its max 5 calls, so it seems like it makes sense. Plus it lets us handle each call individually in the case of an error.

tasks/telemetry.yml
schneideradam: Good point. I think I was using this to view the response. I'll remove.

tasks/pkg-debian/install-apm-inject.yml
schneideradam: Done

tasks/pkg-debian/install-apm-inject.yml
schneideradam: Done

tasks/pkg-redhat/install-apm-inject.yml
schneideradam: Done

tasks/telemetry.yml
schneideradam: Done

tasks/telemetry.yml
schneideradam: Done

README.md
schneideradam: I've added a deprecation warning and also a check that if a user has set the languages value AND NOT set the libraries variable, it will use the deprecated variable.

schneideradam:
@bkabrda  - I've addressed all of your comments. Please take a look and let me know your thoughts. Thanks.

tasks/apm-inject-check.yml
```diff
@@ -76,3 +76,10 @@
 - name: Update datadog_config including config values needed for APM host injection
   set_fact:
     agent_datadog_config: "{{ agent_datadog_config | combine(agent_dd_apm_host_inject_config, list_merge='keep') }}"
+
+- name: Alert for deprecated injection variable
+  debug:
```
bkabrda: It would be better to make this more prominent, because a debug print can easily be missed by users. Can you use [this approach](https://stackoverflow.com/a/63557091)?

tasks/pkg-redhat/install-apm-inject.yml
```diff
@@ -0,0 +1,44 @@
...
+- name: Install APM inject libraries (yum)
+  yum:
+    name: "{{ (item | default([], true)) }}"
+    state: "{{ 'present' if '-' in item else 'latest' }}" # noqa package-latest
+  loop: "{{ agent_dd_apm_install_pkgs }}" # noqa package-latest
+  when: not ansible_check_mode and ansible_pkg_mgr == "yum"
+# End Yum Handlers ----------------------------------
+
+- name: Prevent libraries from being upgraded
+  dpkg_selections:
```
bkabrda: `dpkg_selections` won't work for RPM-based distros. Please take a look at the install_script example and also at `tasks/pkg-redhat.yml` (look for `includepkgs`) - basically we will have to amend the `includepkgs` at that file with all the libraries that are to be pinned.

tasks/pkg-debian/install-apm-inject.yml
```diff
@@ -0,0 +1,30 @@
...
+
+- name: Show installed tracers
+  debug:
+    msg: "Installed tracers: {{ agent_dd_apm_install_pkgs }}"
+
+- name: Prevent libraries from being upgraded
+  dpkg_selections:
+    name: "{{ item.split(':')[0] }}"
+    selection: hold
+  loop: "{{ pre_processed_datadog_apm_libraries }}"
```
bkabrda: Let's use `when: not ansible_check_mode` here as well.

tasks/pkg-debian/install-apm-inject.yml
```diff
@@ -0,0 +1,30 @@
...
+    agent_dd_apm_install_pkgs: "{{ (agent_dd_apm_install_pkgs |
+      default([], true)) + ['datadog-apm-library-' + (item | regex_replace('[:]', '='))] }}"
+  loop: "{{ pre_processed_datadog_apm_libraries }}"
+
+- name: Install APM Inject Library
+  apt:
+    name: 'datadog-apm-inject'
+    state: latest # noqa package-latest
+    update_cache: true
+    cache_valid_time: "{{ datadog_apt_cache_valid_time }}"
```
bkabrda, line 21: Let's use `when: not ansible_check_mode` here as well.

tasks/pkg-debian/install-apm-inject.yml
bkabrda: Did you actually test this? It doesn't work for me, saying:
```
TASK [ansible-datadog : Prevent libraries from being upgraded] ***************************************************
failed: [localhost] (item=python) => {"ansible_loop_var": "item", "changed": false, "item": "python", "msg": "Failed to find package 'python' to perform selection 'hold'."}
```

when I use `datadog_apm_instrumentation_libraries: ["python"]`. Also, I think we only want to use `hold` when a specific version is provided by the user, but not when we're installing latest (in that case I think we want to leave it unpinned, as we do in the install_script).

bkabrda (changes requested):
Hi :wave: thanks for the additional fixes. The PR looks much better now, but there are still places to fix/improve on the newly added functionality.

tasks/pkg-debian/install-apm-inject.yml
schneideradam: Done

tasks/pkg-redhat/install-apm-inject.yml
schneideradam: @bkabrda I am duplicating the function to parse the agent libraries to install here for dnf because i am not sure if this will run in the other call at line 82 of tasks/pkg-redhat.yml. Let me know if that is correct

tasks/telemetry.yml
schneideradam: done

tasks/pkg-redhat/install-apm-inject.yml
schneideradam: done

tasks/apm-inject-check.yml
schneideradam: done

tasks/pkg-debian/install-apm-inject.yml
schneideradam: Fixed this. I realize I was using the wrong variable.

schneideradam:
@bkabrda Ready for review again. Resolved all of your comments and re-tested in both debian and rhel environments.

tasks/apm-inject-check.yml
schneideradam: @bkabrda  Not sure what to do here. My update works as expected, but its failing the linter check. How should we proceed

tasks/pkg-redhat/install-apm-inject.yml
bkabrda: I think it is, but I'll test locally to be 100 % sure. Nice job!

bkabrda:
@schneideradam nice work, I think the code looks really good now. You will need to make sure that the CI is fixed, however. In the meanwhile (some of the failures are very likely related to certain CI fixes that we've done recently, so feel free to rebase on the latest `main` branch, that should help a lot).

I'm going to do some local testing and play around with the feature a little bit to see if everything is 100 % and let you know if I find more places to improve.

schneideradam:
@bkabrda Thanks. I think the main failure is coming from the guidance you gave to modify the deprecation alert message to include fail, when, and ignore_errors. I think the linter doesn't like those. https://github.com/DataDog/ansible-datadog/pull/541#discussion_r1495395668. Not sure how to fix this as its what you guided to doing. It seems I could write a custom module but that seems like overkill. Thoughts?

bkabrda:
@schneideradam please use `failed_when: false` instead of `ignore_errors: true`. That should fix it.

bkabrda:
@schneideradam some tests in the CI are still failing. For example, I'm seeing this error:

```
fatal: [127.0.0.1]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'agent_dd_apm_install_pkgs' is undefined\n\nThe error appears to be in '/root/project/tasks/pkg-redhat.yml': line 96, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n    - name: Set plain includepkgs variable\n      ^ here\n"}
```

Can you please make sure CI is green?

schneideradam:
@bkabrda CI Passing. Let me know if there's anything else here.

bkabrda:
Hi @schneideradam, I did some more testing and found some issues with the pinning logic. My reproducer is:

* Pin version (non-latest) of a specific tracer
* Run Ansible to get it installed
* Now unpin the version to get it upgraded to latest
* Run Ansible again

On Debian/Ubuntu, this errors out:
```
failed: [localhost] (item=datadog-apm-library-python) => {"ansible_loop_var": "item", "cache_update_time": 1678760944, "cache_updated": false, "changed": false, "item": "datadog-apm-library-python", "msg": "'/usr/bin/apt-get -y -o \"Dpkg::Options::=--force-confdef\" -o \"Dpkg::Options::=--force-confold\"       install 'datadog-apm-library-python=2.6.5-1'' failed: E: Held packages were changed and -y was used without --allow-change-held-packages.\n", "rc": 100, "stderr": "E: Held packages were changed and -y was used without --allow-change-held-packages.\n", "stderr_lines": ["E: Held packages were changed and -y was used without --allow-change-held-packages."], "stdout": "Reading package lists...\nBuilding dependency tree...\nReading state information...\nThe following held packages will be changed:\n  datadog-apm-library-python\nThe following packages will be upgraded:\n  datadog-apm-library-python\n1 upgraded, 0 newly installed, 0 to remove and 163 not upgraded.\n", "stdout_lines": ["Reading package lists...", "Building dependency tree...", "Reading state information...", "The following held packages will be changed:", "  datadog-apm-library-python", "The following packages will be upgraded:", "  datadog-apm-library-python", "1 upgraded, 0 newly installed, 0 to remove and 163 not upgraded."]}
```
On CentOS/RHEL, this doesn't fail, but only upgrades the unpinned tracer after Ansible is executed twice in a row (the first time the repofile gets updated *after* the installation, on the second pass this allows yum/dnf to actually upgrade).


For Debian, you should follow the install_script logic, which explicitly unpins first, then installs and then pins. For CentOS/RHEL, you should write the repofile first before doing any install/upgrade, because that will allow the upgrade to happen correctly.

I think this is the last bit that we're missing. Thanks!

bkabrda:
@schneideradam unfortunately there are still issues with the new logic on both Ubuntu and CentOS, let me explain:

* On Ubuntu, you can't use dpkg_selections *before* the package is installed locally, so this fails when first installing packages. I think probably the easiest thing we can do here is use `failed_when: false` on the task (the other way would be to do this only for packages that already have an entry in the `dpkg --get-selections` output, but that's probably too overcomplicated).
* On CentOS/RHEL, the packages don't get pinned - running a `dnf update` still updates them. The reason is that the `includepkgs` line looks something like this: `includepkgs = datadog-agent datadog-apm* datadog-apm-library-js-4.11.1-1 datadog-apm-library-ruby-1.20.0-1` - the `datadog-apm*` is extraneous in this case and will override the pin that comes later. In case *any* tracer packages are explicitly enumerated, we should not use this pin, but rather list them instead of using the `datadog-apm*` glob (and not forget to add datadog-apm-inject).

Does that make sense? Sorry that this is dragging for so long, but we need to get this right to ensure good customer experience.

schneideradam:
Hey, sorry I wasn't quite ready for a review. I was testing on a remote
machine and I had to submit commits to send it over. I'll ping you when its
ready for review. I'm looking into the issues, and yes I think there's a
regression where RHEL is no longer grabbing the package version. I'll get
this fixed up and let you know when ready.  Sorry for all of the back and
forth.

schneideradam:
@bkabrda This is now ready for review. I have tested on both RHEL and Debian.

bkabrda (approved):
Nice job, AFAICS everything is working as we expect it to right now, so I'm approving and merging the PR. Thanks!

PR #554, opened 2024-02-22
Pin Python to version 3.11 in macOS runners

bkabrda:
Pin Python 3.11 which supports all Ansible versions we're testing with.

_Edit - this no longer applies, because some Ansible tested versions don't work with Python 3.12 which ships this pip version:_ We're now getting pip > 23.0.01 installed in macOS runners. This version now requires us to use the `--break-system-packages` switch to mix the pip-installed packages with system packages. We're fine with this because this is only a disposable CI environment.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -200,7 +200,7 @@ jobs:
           command: brew install python3
       - run:
           name: Install Ansible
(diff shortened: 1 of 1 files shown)


amenasria (approved):
LGTM !

PR #556, opened 2024-03-07
Limit the versions of the agent that can be installed on Centos < 7

alopezz:
The approach taken here was:
- When not pinned, install the latest datadog agent version known to work.
- When pinned, error out if the pinned version is above 7.51.

### DIFF
--- a/tasks/os-check.yml
+++ b/tasks/os-check.yml
@@ -3,3 +3,10 @@
   fail:
     msg: "The Datadog Ansible role does not support your OS yet. Please email support@datadoghq.com to open a feature request."
   when: ansible_facts.os_family not in ["RedHat", "Rocky", "AlmaLinux", "Debian", "Suse", "Windows", "Darwin"]
+
+- name: Upper bound last supported Agent version for Centos < 7
+  set_fact:
+    datadog_agent_max_version:
+      full: "{{ datadog_agent_major_version | default('7', true) }}.51.1"
+      minor: 51
+  when: ansible_facts.os_family == "RedHat" and (ansible_facts.distribution_major_version | int) < 7
--- a/tasks/os-check.yml
+++ b/tasks/os-check.yml
@@ -15,6 +15,5 @@
     - name: Set upper agent version bound as ansible fact
       set_fact:
         datadog_agent_max_version:
-          full: "{{ datadog_agent_major_version | default('7', true) }}.51.1"
           minor: 51
       when: rhel_version.stdout | int < 7
--- a/tasks/os-check.yml
+++ b/tasks/os-check.yml
@@ -5,8 +5,16 @@
   when: ansible_facts.os_family not in ["RedHat", "Rocky", "AlmaLinux", "Debian", "Suse", "Windows", "Darwin"]
 
 - name: Upper bound last supported Agent version for Centos < 7
-  set_fact:
-    datadog_agent_max_version:
-      full: "{{ datadog_agent_major_version | default('7', true) }}.51.1"
-      minor: 51
-  when: ansible_facts.os_family == "RedHat" and (ansible_facts.distribution_major_version | int) < 7
+  when: ansible_facts.os_family == "RedHat"
+  block:
+    - name: Get RHEL major version equivalent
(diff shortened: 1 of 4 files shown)


tasks/os-check.yml
```diff
@@ -3,3 +3,18 @@
...
+  block:
+    - name: Get RHEL major version equivalent
+      command: "rpm -E %{rhel}"  # noqa: command-instead-of-module
+      register: rhel_version
+      changed_when: false
+      check_mode: false
+    - name: Set upper agent version bound as ansible fact
+      set_fact:
+        datadog_agent_max_version:
+          full: "{{ datadog_agent_major_version | default('7', true) }}.51.1"
```
spencergilbert: **❓ question:** ‏Do we want to make this work, even if we end up releasing another patch version?

alopezz: I don't see a way to signal to yum that we want `< 7.52`; so this is the next best thing that I've been able to come up with. Are you thinking that we should perhaps just error out in this case?

spencergilbert: Not sure, I think it's better that we follow the behavior we set in the install script - but I don't want us to forget to bump this and leave a vulnerability sitting (for example). It's probably fine as-is, but if someone knows of a way to do `<.52` that feels better.

tasks/os-check.yml
```diff
@@ -3,3 +3,18 @@
   fail:
     msg: "The Datadog Ansible role does not support your OS yet. Please email support@datadoghq.com to open a feature request."
   when: ansible_facts.os_family not in ["RedHat", "Rocky", "AlmaLinux", "Debian", "Suse", "Windows", "Darwin"]
+
+- name: Upper bound last supported Agent version for Centos < 7
+  when: ansible_facts.os_family == "RedHat"
+  block:
+    - name: Get RHEL major version equivalent
+      command: "rpm -E %{rhel}"  # noqa: command-instead-of-module
```
bkabrda: Hmm, this is interesting, I never thought about doing this. Have you tested on e.g. Scientific Linux 6? I feel like there might be rhel-6 based distros where this macro wouldn't expand, but I'm not sure.

tasks/os-check.yml
bkabrda: I think specifying version range should work, but you'll have to use the epoch number (`1:`) which the Agent packages have had since the beginning.

I think something like `yum install "datadog-agent < 1:7.52.0-1"` should work, but you'd have to figure out how best to pass it to the ansible tasks which might be doing additional parsing etc.

alopezz: I'll give that a shot.

tasks/os-check.yml
alopezz: I did a quick test on a Scientific Linux 6 box and it did expand as expected (to 6). Would you say it's safer to use something else (or at least to have some sort of fallback)?

tasks/os-check.yml
alopezz: This worked well 👍

tasks/pkg-redhat/install-latest.yml
```diff
@@ -1,7 +1,16 @@
 ---
+- name: Set installation target for latest
+  set_fact:
+    # Apply a version cap if required, otherwise just target the agent package
+    datadog_agent_target: >-
+      {{ datadog_agent_flavor }}
+      {% if datadog_agent_max_minor_version is defined %}
+       < 1:{{ agent_datadog_agent_major_version }}.{{ datadog_agent_max_minor_version + 1 }}.0-1
```
spencergilbert, line 8: **❓ question:** ‏ Wouldn't this always end up targeting `7.52.0`?

alopezz: It might not be obvious with all the templating, but this actually sets the target to `"datadog-agent < 1:7.52.0-1"` as suggested by @bkabrda above. In my local tests on Centos 6 this results in the installation of 7.51.1 which is what we want.

spencergilbert: Oh, as you say that now it seems more obvious 👍

spencergilbert (approved):

PR #561, opened 2024-03-29
Install expired 2022 rpm gpg key only when needed on agent <= 7.35

Pythyu:
Only download and install expired 2022 RPM gpg key on agent <= 7.35

### DIFF
--- a/tasks/pkg-redhat.yml
+++ b/tasks/pkg-redhat.yml
@@ -64,12 +64,13 @@
         dest: /tmp/DATADOG_RPM_KEY_E09422B3.public
         mode: '600'
         checksum: "sha256:{{ datadog_yum_gpgkey_e09422b3_sha256sum }}"
+      when: agent_datadog_minor | int < 36
 
     - name: Import new RPM key (Expires in 2022)
       rpm_key:
         key: /tmp/DATADOG_RPM_KEY_E09422B3.public
         state: present
-      when: not ansible_check_mode
+      when: not ansible_check_mode and agent_datadog_minor | int < 36
 
     - name: Download new RPM key (Expires in 2024)
       get_url:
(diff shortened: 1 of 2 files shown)


tasks/pkg-redhat.yml
```diff
@@ -64,12 +64,13 @@
         dest: /tmp/DATADOG_RPM_KEY_E09422B3.public
```
chouquette, line 64: I can't comment on line 61, but we might want to stop referring to a key expired 2 years ago as `new` :sweat_smile:

tasks/pkg-redhat.yml
```diff
@@ -64,12 +64,13 @@
...
         mode: '600'
         checksum: "sha256:{{ datadog_yum_gpgkey_e09422b3_sha256sum }}"
+      when: agent_datadog_minor is defined and agent_datadog_minor | int < 36
 
     - name: Import new RPM key (Expires in 2022)
```
chouquette: same remark about "new" here

tasks/pkg-redhat.yml
```diff
@@ -64,12 +64,13 @@
...
       rpm_key:
         key: /tmp/DATADOG_RPM_KEY_E09422B3.public
         state: present
-      when: not ansible_check_mode
+      when: not ansible_check_mode and agent_datadog_minor is defined and agent_datadog_minor | int < 36
```
chouquette, line 73: :question: Isn't `agent_datadog_minor` always defined? Or if it can be undefined, what's the expected context?

Or to ask it another way, this appears to invert the logic from the install script (never install the key when minor version isn't set), is that the expected logic?

Pythyu: `agent_datadog_minor` can be undefined, meaning install the latest version

chouquette: I might have not been awake enough when I initially reviewed, the logic appears the same :thinking:

chouquette (approved):

PR #566, opened 2024-04-16
Remove `warn` argument that is no longer available in newer Ansible

### LINKED ISSUE #564: Role not compatible with Ansible versions higher than v6.7.0
## Problem
I'm currently blocked to use the latest version of the role to install the Datadog Agent on machines with SLES 12 SP5 when using any Ansible version higher than v6.7.0 (Ansible-core 2.13).

### Reason
The task in https://github.com/DataDog/ansible-datadog/blob/4.22.0/tasks/pkg-suse.yml#L133-L139 and perhaps more rely on the deprecated and in v7 and higher removed feature to mute Ansible warnings. 

See Ansible 7 porting guide: 
* https://github.com/ansible/ansible-documentation/blob/devel/docs/docsite/rst/porting_guides/porting_guide_7.rst?plain=1#L940

### Affected versions
* Datadog role **datadog.datadog 4.21.0 / 2023-12-04**

--- 

I hope this is detailed enough for you to fix the issue. Thanks

alopezz:
Fixes #564. Any execution path that hit any of the affected commands would fail on Ansible 7 and above.

Further context can be found in https://github.com/ansible/ansible/blob/v2.11.0/changelogs/CHANGELOG-v2.11.rst#deprecated-features.

I don't foresee any harm, other than possibly users on very old Ansible versions starting to see a warning about the existence of a module that could be used instead of some explicit command.

Note: I've tested this manually. I intend to follow this up with adding newer Ansible versions to the test matrices that we have, but that's gonna take a little bit more time and I'd rather put out this fix as soon as we can.

### DIFF
--- a/tasks/pkg-redhat.yml
+++ b/tasks/pkg-redhat.yml
@@ -202,22 +202,18 @@
       register: agent_repofilecustom
       when: (datadog_yum_repo | length > 0) and (not ansible_check_mode)
 
-    - name: Clean repo metadata if repo changed # noqa no-handler args[module] (args.warn is deprecated since Ansible 2.11)
+    - name: Clean repo metadata if repo changed # noqa: command-instead-of-module
       command: yum clean metadata --disablerepo="*" --enablerepo=datadog
       failed_when: false # Cleaning the metadata is only needed when downgrading a major version of the Agent, don't fail because of this
-      args:
-        warn: false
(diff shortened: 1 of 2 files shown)


Pythyu (approved):
LGTM a follow-up update will be needed on the [ansible-collection](https://github.com/ansible-collections/Datadog) repository to check that the collection still function with the latest role

aairey (approved):

PR #568, opened 2024-05-21
add new future GPG key following 2024 GPG key rotation

Pythyu:
add new future GPG keys for RPM and APT following the 2024 GPG key rotation


f4usto (approved):
lgtm

@Pythyu, can we collect all the PRs done for this effort somehow?

PR #573, opened 2024-06-21
add support for datadog-installer

chouquette:
This PR adds initial support for the datadog-installer.
It will install it if APM is enabled, as is the case for the install script, and will delegate installation of the injector and tracer libraries to it. 
Installation of tracer libraries are now gated behind a check that ensures the installer doesn't currently own that package.
It also provides telemetry similar to what the install script provides
BARX-381

### DIFF
--- a/tasks/installer-setup.yml
+++ b/tasks/installer-setup.yml
@@ -7,6 +7,7 @@
 - name: Start duration measurements
   command: "date +%s%N"
   register: datadog_installer_start_time
+  changed_when: True
 
 - name: Install datadog-installer package (dnf)
   dnf:
@@ -53,12 +54,14 @@
     DD_APM_INSTRUMENTATION_LANGUAGES: "{{ datadog_apm_instrumentation_libraries | join(',') }}"
   ignore_errors: true
   when: not datadog_installer_install_result.failed
+  changed_when: True
 
 - name: Check if installer owns datadog-agent package
-  shell: datadog-installer is-installed "{{ datadog_agent_flavor }}"
+  command: datadog-installer is-installed "{{ datadog_agent_flavor }}"
   failed_when: datadog_installer_owns_agent.rc != 0 and datadog_installer_owns_agent.rc != 10
   register: datadog_installer_owns_agent
   when: not datadog_installer_bootstrap_result.failed
+  changed_when: True
 
 - name: Disable agent install if owned by installer
   set_fact:
@@ -66,11 +69,12 @@
   when: datadog_installer_owns_agent.rc == 0
 
 - name: Query APM packages owned by installer
-  shell: datadog-installer is-installed "datadog-apm-library-{{ item }}"
+  command: datadog-installer is-installed "datadog-apm-library-{{ item }}"
   register: datadog_installer_owned_apm_packages
   loop: "{{ datadog_apm_instrumentation_libraries }}"
   when: not datadog_installer_bootstrap_result.failed
   failed_when: datadog_installer_owned_apm_packages.rc != 0 and datadog_installer_owned_apm_packages.rc != 10
+  changed_when: True
   environment:
     DD_API_KEY: "{{ datadog_api_key }}"
 
--- a/tasks/installer-setup.yml
+++ b/tasks/installer-setup.yml
@@ -124,7 +124,7 @@
     headers:
       DD-Api-Key: "{{ datadog_api_key }}"
     body_format: json
-  ignore_errors: true
+  failed_when: false
 
 - name: Setup logs body
   set_fact:
@@ -155,4 +155,4 @@
     headers:
       DD-Api-Key: "{{ datadog_api_key }}"
     body_format: json
-  ignore_errors: true
+  failed_when: false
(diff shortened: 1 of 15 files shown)


hestonhoffman (approved):

tasks/installer-setup.yml
```diff
@@ -0,0 +1,158 @@
...
+  command: /usr/bin/datadog-bootstrap bootstrap
+  register: datadog_installer_bootstrap_result
+  environment:
+    DATADOG_TRACE_ID: "{{ datadog_installer_trace_id }}"
+    DATADOG_PARENT_ID: "{{ datadog_installer_trace_id }}"
+    DD_SITE: "{{ datadog_site | default('datadoghq.com') }}"
+    DD_API_KEY: "{{ datadog_api_key }}"
+    DD_REMOTE_UPDATES: "{{ 'true' if datadog_remote_updates is defined and datadog_remote_updates else '' }}"
+    DD_APM_INSTRUMENTATION_ENABLED: "{{ datadog_apm_instrumentation_enabled }}"
+    DD_APM_INSTRUMENTATION_LANGUAGES: "{{ datadog_apm_instrumentation_libraries | join(',') }}"
```
BaptisteFoy: We have `DD_APM_INSTRUMENTATION_LIBRARIES` in the installer, I'm not sure but shouldn't we align? cc @paullegranddc

tasks/installer-setup.yml
```diff
@@ -0,0 +1,158 @@
...
+
+- name: Save APM packages before filtering
+  set_fact:
+    datadog_apm_instrumentation_libraries_unfiltered: "{{ datadog_apm_instrumentation_libraries }}"
+
+- name: Filter APM packages owned by the installer
+  set_fact:
+    datadog_apm_instrumentation_libraries: "{{ datadog_apm_instrumentation_libraries | difference([item.item]) }}"
+  when: item.rc == 0
+  loop: "{{ datadog_installer_owned_apm_packages.results }}"
```
BaptisteFoy, line 97: Does this work when the var isn't registered or should we add `when: not datadog_installer_bootstrap_result.failed`?

tasks/installer-setup.yml
```diff
@@ -0,0 +1,158 @@
...
+  when: ansible_facts.os_family in ["RedHat", "Rocky", "AlmaLinux"] and not ansible_check_mode and ansible_pkg_mgr == "dnf"
+
+- name: Install latest datadog-agent package (yum)
+  yum:
+    name: "datadog-installer"
+    update_cache: true
+    state: latest # noqa package-latest
+  register: datadog_installer_install_result
+  ignore_errors: true
+  when: ansible_facts.os_family in ["RedHat", "Rocky", "AlmaLinux"] and not ansible_check_mode and ansible_pkg_mgr == "yum"
```
BaptisteFoy, line 31: We don't support CentOS 6, should we add a check like we did in the install script like [we did in the install script](https://github.com/DataDog/agent-linux-install-script/pull/212)?

Pythyu (commented):
Looks good, if it doesn't add too much time to the PR adding/modify a testinstall to check for the installer installation would be great

tasks/installer-setup.yml
chouquette: Let's be on the safer side, I added the clause

tasks/installer-setup.yml
paullegranddc: The install-script reads both for backward compatibility purpose (actually we should too maybe?) but `DD_APM_INSTRUMENTATION_LANGUAGES` is legacy, so this should be `DD_APM_INSTRUMENTATION_LIBRARIES`

Pythyu (approved):

raphaelgavache:
I'm not familiar enough with Ansible and I couldn't from the PR reading assert this:

[APM install scripts ](https://github.com/DataDog/ansible-datadog/blob/8da46f0aa365d7ff4955fe9942699d99a56f1114/tasks/apm-inject-install.yml#L10-L43) need to be gated on[ apm-inject installation by the injector](https://github.com/DataDog/agent-linux-install-script/blob/main/install_script.sh.template#L1806)

If the injector is installing apm-inject the scripts shouldn't run.

chouquette:
> I'm not familiar enough with Ansible and I couldn't from the PR reading assert this:
> 
> [APM install scripts ](https://github.com/DataDog/ansible-datadog/blob/8da46f0aa365d7ff4955fe9942699d99a56f1114/tasks/apm-inject-install.yml#L10-L43) need to be gated on[ apm-inject installation by the injector](https://github.com/DataDog/agent-linux-install-script/blob/main/install_script.sh.template#L1806)
> 
> If the injector is installing apm-inject the scripts shouldn't run.

Thanks for spotting! This should now be fixed
I think the entirety of `apm-inject-install.yml` should be disabled as it also includes some package pinning and listing what's installed by apt/rpm, but we might want to introduce something similar to list what the installer installed.

chouquette:
> I think the entirety of apm-inject-install.yml should be disabled

The CI begs to differ :sweat_smile:

chouquette:
It works better when installing the tracers that are not handled by the installer, who'd have thought :facepalm:

tasks/installer-setup.yml
chouquette: done here: https://github.com/DataDog/ansible-datadog/pull/573/files#diff-628b17105eb5471592a85311eb58449633695b482221d729f7e3965cd0ba21cbR14-R17

PR #584, opened 2024-07-16
Speed up the conf.d checks

a-rhodes:
When using the config options `datadog_disable_untracked_checks`, `datadog_disable_default_checks` or `datadog_disable_example_checks`, the Ansible run is very slow as it runs a separate task for every directory in the `conf.d` folder.

Instead, if we change the `find` call to look for conf.yaml* files in the subdirectories, we can filter the results and target specifically what we need, resulting in a much faster role.

### DIFF
--- a/tasks/_agent-linux-macos-shared.yml
+++ b/tasks/_agent-linux-macos-shared.yml
@@ -18,20 +18,26 @@
   when: datadog_manage_config
   notify: "{{ agent_dd_notify_agent }}"
 
-- name: Register all checks directories present in datadog
+- name: Register all check configs present in datadog
   find:
     paths: "{{ agent_dd_config_dir }}/conf.d/"
     patterns:
-      - "*.d"
-    file_type: directory
-  register: agent_datadog_conf_directories
+      - "conf.yaml*"
+    recurse: true
+    depth: 2
+    file_type: file
+  register: agent_datadog_conf_paths
   when: datadog_manage_config and (datadog_disable_untracked_checks or datadog_disable_default_checks)
 
 - name: Delete checks not present in agent_datadog_tracked_checks
   file:
     path: "{{ agent_dd_config_dir }}/conf.d/{{ item }}.d/conf.yaml"
     state: absent
-  loop: "{{ agent_datadog_conf_directories.files | map(attribute='path') | list | map('basename') | list | map('regex_replace', '^(.*).d$', '\\1') | list }}"
+  loop: >-
+    {{
+      agent_datadog_conf_paths.files | map(attribute='path') | select('match', '.*\/conf\.yaml$') | list
+      | map('dirname') | map('basename') | list | map('regex_replace', '^(.*).d$', '\1') | list
+    }}
   when: datadog_manage_config and datadog_disable_untracked_checks and item not in agent_datadog_tracked_checks
   notify: "{{ agent_dd_notify_agent }}"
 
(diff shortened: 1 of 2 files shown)


tasks/_agent-linux-macos-shared.yml
```diff
@@ -18,36 +18,41 @@
...
+    file_type: file
+  register: agent_datadog_conf_paths
   when: datadog_manage_config and (datadog_disable_untracked_checks or datadog_disable_default_checks)
 
 - name: Delete checks not present in agent_datadog_tracked_checks
   file:
     path: "{{ agent_dd_config_dir }}/conf.d/{{ item }}.d/conf.yaml"
     state: absent
-  loop: "{{ agent_datadog_conf_directories.files | map(attribute='path') | list | map('basename') | list | map('regex_replace', '^(.*).d$', '\\1') | list }}"
+  loop: "{{ agent_datadog_conf_paths.files | selectattr('filename', 'eq', 'conf.yaml') | map(attribute='path') | list
```
alopezz: It looks like the `filename` attribute doesn't get defined. Following the docs on [find](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/find_module.html#return-files), the list of attributes would seem to be the one on https://docs.ansible.com/ansible/latest/collections/ansible/builtin/stat_module.html#return-stat; I don't see `filename` there. So I'm not sure this can work like that.

alopezz (changes requested):
👋 Hi, thanks for the PR!

I've tried to test it against a Ubuntu VM, setting `datadog_disable_untracked_checks: true` and I get this error:

```
TASK [datadog.datadog : Delete checks not present in agent_datadog_tracked_checks] *****
fatal: [default]: FAILED! => {"msg": "'dict object' has no attribute 'filename'"}
```

tasks/_agent-linux-macos-shared.yml
a-rhodes: My apologies, I tested on a Windows host, and just ported this over from that code. I didn't realize `find` returned different attributes than `win_find`. I'll get this work on linux and post an update.

a-rhodes: Should be all set now. I removed the filename dependency all together and just match the value of path instead

tasks/_agent-linux-macos-shared.yml
```diff
@@ -18,36 +18,50 @@
...
 - name: Delete checks not present in agent_datadog_tracked_checks
   file:
     path: "{{ agent_dd_config_dir }}/conf.d/{{ item }}.d/conf.yaml"
     state: absent
-  loop: "{{ agent_datadog_conf_directories.files | map(attribute='path') | list | map('basename') | list | map('regex_replace', '^(.*).d$', '\\1') | list }}"
+  loop: >-
+    {{
+      agent_datadog_conf_paths.files | map(attribute='path') | select('match', '.*\/conf\.yaml$') | list
+      | map('dirname') | map('basename') | list | map('regex_replace', '^(.*).d$', '\1') | list
+    }}
```
alopezz, line 40: 💭 How about something like
```suggestion
    {{ agent_datadog_conf_paths.files | map(attribute='path') | select('match', '.*\/conf\.yaml$') | list }}
```

And replacing `path` to be:
```yaml
    path: "{{ item }}"
```
(and similarly for the others)?

That should do the same and allow us to simplify this filter chain significantly.

alopezz (approved):
This looks good now and is indeed much faster. Thanks!

I do have a suggestion that could simplify the code while maintaining the enhancement, let me know what you think.

tasks/_agent-linux-macos-shared.yml
a-rhodes: Works for me. It does change the Ansible task output a bit since the value of item is different. But I agree this is a much cleaner option.

a-rhodes: Just kidding, I remember when I didn't do it like that. `item` needs to be the check name so it can be used in the `when` statement

alopezz: Good point, I had missed that. I'll merge this as is, thanks again.

PR #590, opened 2024-08-01
[FA] Fix APM inject config with the installer

coignetp:
We shouldn't need this config with the installer


chouquette (approved):
LGTM, if you want to rebase it on top of main as it doesn't require the changes in `chouquette/handle_agent_installed_installer` please feel free to do so, otherwise I'll merge it afterward

PR #591, opened 2024-08-01
Fix security-agent.yaml.j2

jcarroll-ls:
Missing line break seems to cause the `compliance_config` variable to be indented and therefore placed within the `runtime_security_config` variable which leads to an incorrect configuration for CSM.


AliDatadog (approved):

AliDatadog:
Thank you for your contribution !

PR #594, opened 2024-08-08
Fix datadog-installer handling on SUSE

chouquette:
I completely overlooked the existing differences between SUSE and RHEL, causing the installer setup to be missing for that distribution

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -310,7 +310,7 @@ jobs:
           -e datadog_remote_updates="true"
       - run: >
           bash -c 'datadog-installer version;
-              if [ ! -d /opt/datadog-packages/agent ]; then
+              if [ ! -d /opt/datadog-packages/datadog-agent ]; then
                 echo "The agent should have been installed by the installer";
                 exit 1;
               fi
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -284,6 +284,37 @@ jobs:
...
+      - run: >
+          ANSIBLE_JINJA2_NATIVE="<<parameters.jinja2_native>>" ansible-playbook
+          -i ./ci_test/inventory/<<parameters.inventory>> "./ci_test/install_installer.yaml"
+          -e datadog_remote_updates="true"
+      - run: >
+          bash -c 'datadog-installer version;
+              if [ -d /opt/datadog-agent ]; then
+                echo "The agent should NOT have been installed by the distribution";
+                exit 1;
+              fi'
```
f4usto: This is testing the negative, should we test that the agent is installed on the installer expected path?

chouquette: I added an additional conditional for the presence of the datadog-packages folder.

FWIW, we'll soon enable APM on SUSE through ansible, which will make this special case unneeded

spencergilbert (approved):

f4usto (approved):

PR #602, opened 2024-08-27
add ansible_managed to checks.yaml template

rockaut:
this simply adds the ansible_managed line to the checks.yaml template
ansible_managed is in every other meaningful settings template included just not in the check config template


chouquette (approved):
LGTM, thanks!

PR #609, opened 2024-09-10
Fix installation when a version is pinned and the installer is enabled

chouquette:


.circleci/config.yml
```diff
@@ -453,3 +469,6 @@ workflows:
          matrix:
            parameters:
               ansible_version: ["2_10", "3_4", "4_10"]
+
+      - test_installer_over_pinned
```
Pythyu, line 473: 🐒 Nitpick :
Looks like a sed error

tasks/main.yml
```diff
@@ -35,7 +35,7 @@
 
 - name: Debian Install Tasks
   include_tasks: pkg-debian.yml
-  when: ansible_facts.os_family == "Debian" and not agent_datadog_skip_install
```
Pythyu, line 38: ❓ Question :
Are we still using `agent_datadog_skip_install` or it's unsupported from this PR on ?

Pythyu (approved):
Otherwise LGTM 👍

tasks/main.yml
chouquette: It's still read from down the line, and is used by the installer to disable installing the agent through the package manager

.circleci/config.yml
chouquette: It's adding the job defined above to the workflow :grin: 

I don't think there's a way to define and add the job as a single step, or at least I haven't found one :(

Pythyu: Aaah I see didn't knew we could call it like that 👍

PR #613, opened 2024-09-19
installer: don't crash when bootstrap fails

chouquette:
When the bootstrap operation fails, we don't run the task that registers the `datadog_installer_owns_agent` facts, which causes the role to crash when accessing it

### DIFF
--- a/tasks/installer-setup.yml
+++ b/tasks/installer-setup.yml
@@ -75,7 +75,7 @@
 - name: Disable agent install if owned by installer
   set_fact:
     agent_datadog_skip_install: true
-  when: datadog_installer_owns_agent.rc == 0
+  when: not datadog_installer_bootstrap_result.failed and datadog_installer_owns_agent.rc == 0
 
 - name: Query APM packages owned by installer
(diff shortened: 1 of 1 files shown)


tasks/installer-setup.yml
```diff
@@ -75,7 +75,7 @@
 - name: Disable agent install if owned by installer
   set_fact:
     agent_datadog_skip_install: true
-  when: datadog_installer_owns_agent.rc == 0
+  when: not datadog_installer_bootstrap_result.failed and datadog_installer_owns_agent.rc == 0
```
Pythyu, line 78: ❓ Question :
Just by lack of context, what's supposed to happen when the bootstrap fail aside from crashing ? Or would it be safer to crash ? to notify the failure ?

Pythyu (approved):
Otherwise LGTM

tasks/installer-setup.yml
chouquette: We carry on in order to report telemetry afterward, and only then notify of the error.
There might be some more elegant ways to do so though

PR #616, opened 2024-09-20
reintroduce dependabot.yaml

AliDatadog:
Dependabot was removed in https://github.com/DataDog/ansible-datadog/pull/394. This PR reintroduces it.


chouquette (approved):
LGTM, I'd just like to know if @KSerrania has some context around the removal. It appears to have been intentional, but I don't understand why

PR #619, opened 2024-10-14
[AGENT-12513] Fix Ansible check mode when installer/apm injection is enabled

### LINKED ISSUE #610: Breaks when in check mode
When you run the new version with --check it failing due to conditional check as it can not find `datadog_installer_install_result.failed`

This is due to the include tasks to install for either `DNF` , `YUM`, `DEB`, and `zypper` get skipped when in check mode so the var `datadog_installer_install_result.failed` is not defined

tasks/installer-setup.yml ln 15 to 29

has this code 
`- name: Include installer DNF install task
  ansible.builtin.include_tasks: pkg-redhat/install-installer-dnf.yml
  when: ansible_facts.os_family in ["RedHat", "Rocky", "AlmaLinux"] and not ansible_check_mode and ansible_pkg_mgr == "dnf"

- name: Include installer YUM install task
  ansible.builtin.include_tasks: pkg-redhat/install-installer-yum.yml
  when: ansible_facts.os_family in ["RedHat", "Rocky", "AlmaLinux"] and not ansible_check_mode and ansible_pkg_mgr == "yum"

- name: Include installer DEB install task
  ansible.builtin.include_tasks: pkg-debian/install-installer.yml
  when: ansible_facts.os_family == "Debian" and not ansible_check_mode

- name: Include installer Zypper install task
  ansible.builtin.include_tasks: pkg-suse/install-installer-zypper.yml
  when: ansible_facts.os_family == "Suse" and not ansible_check_mode`
  
from what i can tell so far we could just remove the `and not ansible_check_mode` check from those lines and this will allow the check mode to run without error  or we could use the `not datadog_installer_install_result.failed | default(false) ` on the conditional check to stop it throwing a failed action maybe 
  
also this only seem to fail on check mode when enable apm
`    datadog_apm_instrumentation_enabled: "host"
    datadog_apm_instrumentation_libraries: ["all"]`

the test playbook i used to get this error

`- hosts: localhost
  gather_facts: true
  tasks:
    - name: Import Datadog Agent role
      import_role:
        name: datadog.dd.agent
  vars:
    datadog_api_key: "********"
    datadog_site: "datadoghq.eu"
    datadog_apm_instrumentation_enabled: "host"
    datadog_apm_instrumentation_libraries: ["all"]`

if we run this with no --check flag then it will work fine.

Command : `ansible-playbook datadog.yml --diff --check`
Output:
`TASK [datadog.dd.agent : Include installer DNF install task] **********************************************************************************************************************************************
skipping: [127.0.0.1]

TASK [datadog.dd.agent : Include installer YUM install task] **********************************************************************************************************************************************
skipping: [127.0.0.1]

TASK [datadog.dd.agent : Include installer DEB install task] **********************************************************************************************************************************************
skipping: [127.0.0.1]

TASK [datadog.dd.agent : Include installer Zypper install task] *******************************************************************************************************************************************
skipping: [127.0.0.1]

TASK [datadog.dd.agent : Bootstrap the installer] *********************************************************************************************************************************************************
fatal: [127.0.0.1]: FAILED! => {"msg": "The conditional check 'not datadog_installer_install_result.failed' failed. The error was: error while evaluating conditional (not datadog_installer_install_result.failed): 'datadog_installer_install_result' is undefined. 'datadog_installer_install_result' is undefined\n\nThe error appears to be in '/home/****/.ansible/collections/ansible_collections/datadog/dd/roles/agent/tasks/installer-setup.yml': line 31, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Bootstrap the installer\n  ^ here\n"}
...ignoring`

tbavelier:
### What does this PR do ?
* For the APM inject task, move the `not ansible_check_mode` condition to the parent `when`. We should not run any of the commands/file from `dd-host-install` in check mode which we were doing individually (except injecting the Docker config which we should not have done in check mode, so this "fixes" it incidentally)
* For the installer task, the playbook "depends" on `datadog_installer_install_result` which itself is only registered when not running in check mode, as coded in each `tasks/pkg-`, e.g.: https://github.com/DataDog/ansible-datadog/blob/a34e9623b9440c4f4ae11d19be1ac24a320e7936/tasks/pkg-debian/install-installer.yml#L8. Thus, we add `not ansible_check_mode` on all the subsequent steps that depend on this registry, as we should not run any bootstrap/installer command, nor attempt to send any telemetry in check mode
* In the main Agent linux task, do not try to check for services that might not exist (e.g. user never ran without `--check` yet)

### Motivation
Closes https://github.com/DataDog/ansible-datadog/issues/610

### QA instructions
Use a playbook with the installer:
#### Role
```yaml
- hosts: myhosts
  roles:
    - { role: ansible-datadog, become: yes }
  vars:
    datadog_api_key: baguette
    datadog_enabled: false
    datadog_apm_instrumentation_enabled: "host"
    datadog_apm_instrumentation_libraries: ["java:1", "python:2", "js:5", "dotnet:3", "ruby:2"]
```

#### Collection
```yaml
- hosts: myhosts
  tasks:
  - name: Import the Datadog Agent role from the DD collection
    import_role:
      name: datadog.dd.agent
  vars:
    datadog_api_key: baguette
    datadog_enabled: false
    datadog_apm_instrumentation_enabled: "host"
    datadog_apm_instrumentation_libraries: ["java:1", "python:2", "js:5", "dotnet:3", "ruby:2"]
```
Run it in check mode and ensure it does not error out: `ansible-playbook -i inventory.ini playbook.yml --check`

### DIFF
--- a/tasks/installer-setup.yml
+++ b/tasks/installer-setup.yml
@@ -50,38 +50,38 @@
     DD_AGENT_MAJOR_VERSION: "{{ datadog_agent_major_version | default('') }}"
     DD_AGENT_MINOR_VERSION: "{{ datadog_agent_minor_version | default('') }}"
   ignore_errors: true
-  when: not datadog_installer_install_result.failed
+  when: not ansible_check_mode and not datadog_installer_install_result.failed
   changed_when: true
 
 - name: Check if installer owns datadog-agent package
   command: datadog-installer is-installed "{{ datadog_agent_flavor }}"
   failed_when: datadog_installer_owns_agent.rc != 0 and datadog_installer_owns_agent.rc != 10
   register: datadog_installer_owns_agent
-  when: not datadog_installer_bootstrap_result.failed
+  when: not ansible_check_mode and not datadog_installer_bootstrap_result.failed
   changed_when: true
 
 - name: Check if installer owns apm injector package
   command: datadog-installer is-installed "{{ datadog_inject_apm_flavor }}"
   failed_when: datadog_installer_owns_injector_result.rc != 0 and datadog_installer_owns_injector_result.rc != 10
   register: datadog_installer_owns_injector_result
-  when: not datadog_installer_bootstrap_result.failed
+  when: not ansible_check_mode and not datadog_installer_bootstrap_result.failed
   changed_when: true
 
 - name: Register if the installer owns apm injector package
   set_fact:
     datadog_installer_owns_injector: "{{ datadog_installer_owns_injector_result.rc == 0 }}"
-  when: not datadog_installer_bootstrap_result.failed
+  when: not ansible_check_mode and not datadog_installer_bootstrap_result.failed
 
 - name: Disable agent install if owned by installer
   set_fact:
     agent_datadog_skip_install: true
-  when: not datadog_installer_bootstrap_result.failed and datadog_installer_owns_agent.rc == 0
+  when: not ansible_check_mode and not datadog_installer_bootstrap_result.failed and datadog_installer_owns_agent.rc == 0
 
 - name: Query APM packages owned by installer
   command: datadog-installer is-installed "datadog-apm-library-{{ item }}"
   register: datadog_installer_owned_apm_packages
   loop: "{{ datadog_apm_instrumentation_libraries }}"
-  when: not datadog_installer_bootstrap_result.failed
+  when: not ansible_check_mode and not datadog_installer_bootstrap_result.failed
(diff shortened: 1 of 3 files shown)


tasks/installer-setup.yml
chouquette, line 1: Wouldn't it be easier not to include the file when running in check mode? Almost every tasks are disabled and the remaining ones don't seem very valuable to check

chouquette (approved):
One nit/remark. LGTM otherwise

PR #624, opened 2024-11-18
remove deprecated apm deb/rpms

raphaelgavache:
Removes `Install APM inject libraries (dnf)` from redhat steps since it is handled by the installer

Ran the following playbook on RockyLinux vm:
```
- hosts: myhosts
  tasks:
  - name: Print message
    ansible.builtin.debug:
      msg: Hello world
  - name: Import the Datadog Agent role from the DD collection
    import_role:
      name: ansible-datadog
  vars:
    datadog_api_key: <KEY>
    datadog_enabled: true
    datadog_remote_updates: true
    datadog_installer_enabled: true
    datadog_config:
      hostname_trust_uts_namespace: true
    datadog_apm_instrumentation_enabled: host
    datadog_apm_instrumentation_libraries: ["java:1.40.0"]
```
All checks pass


cswatt (approved):

tasks/pkg-redhat.yml
```diff
@@ -122,24 +122,16 @@
...
-      loop: "{{ pre_processed_datadog_apm_libraries | default([], true) }}"
-      when: not ansible_check_mode and datadog_apm_instrumentation_enabled | length > 0
-
     - name: Set versioned includepkgs variable
       set_fact:
         agent_datadog_includepkgs: "{{ datadog_agent_flavor }}-{{ agent_datadog_agent_redhat_version
         | regex_replace('^\\d+:', '') }}
-        {{ datadog_inject_apm_flavor }}
-        {{ agent_dd_apm_install_pkgs | default([], true) | join(' ') }}"
+        {{ datadog_inject_apm_flavor }}"
```
tbavelier:
```suggestion
        "
```
I believe we should remove `datadog_inject_apm_flavor`, it was solely used for deb/rpm injection ? We already remove it in other places. Note the suggestion probably cannot be applied as is for linting

defaults/main.yml
tbavelier, line 1: `datadog_inject_apm_flavor` should also be removed from this file, see other comment

tbavelier (commented):
LGTM overall, I think we can get rid of `datadog_inject_apm_flavor` in a few more places.
Tested successfully on Rockylinux 8 with Ansible core 2.16 (2.17 is not compatible with RHEL 8 if using dnf which we do https://www.reddit.com/r/ansible/comments/1d9pfhf/dnf_redhat8_rocky_8_not_compatible_with)

tbavelier (approved):
macOS CI failures should be un-related

PR #647, opened 2025-05-02
Remove centos 6/7 from readme

swang392:
Documentation for CentOS6/7 is no longer needed in the README as support for it was removed in ansible role 5.0.0


tbavelier (commented):
Maybe we could add a "deprecated" or something to only say this section applies to version of the role < 5.x ? And that means we also most likely have code we can clean up `datadog_ignore_old_centos_python3_error`

drichards-87 (approved):

tbavelier (approved):

PR #648, opened 2025-05-06
fixed typo in fix-parse-version-windows.yml

jacob9423:
Using this module to help install Datadog on new servers. 

Co worker was trying to use it today and found it was giving an error  on "Include Windows Agent Version Tasks". I went though the code and found that in that task it was using the module `amazon.windows.win_shell` it should be `ansible.windows.win_shell` so I fixed that.

### DIFF
--- a/tasks/parse-version-windows.yml
+++ b/tasks/parse-version-windows.yml
@@ -1,7 +1,7 @@
 ---
 # NOTE: This won't work with rc / beta builds.
 - name: Get Windows Agent version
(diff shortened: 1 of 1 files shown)


tbavelier (approved):
Thanks for catching that!

swang392:
Hi @jacob9423, thanks again for catching this! The issue has been fixed in version 5.1.1 of this role. The corresponding collection will be released shortly.

PR #665, opened 2025-08-29
[AGENTONB-2490] Support `remote_updates` through embedded installer

tbavelier:
* Do not install deprecated installer (deb/rpm) if `datadog_remote_updates` is set to `true`: since `7.69`, `remote_updates` simply needs to be in `datadog.yaml` and installer will take over
* During an upgrade from Fleet (`datadog-agent-exp` service is running), when running the playbook, do not restart/ensure Agent service is running, as during the experiment phase, the Agent service is stopped.
* Remove/update tests that would check for installer with remote_updates

For completeness, a test should be added to verify an ansible playbook runs does not restart the service during an upgrade, e.g. by using `datadog-installer install-experiment` to simulate an upgrade but that will be done in a future.
For now, I tested manually and ensured it indeed does not restart during an upgrade, and resumes afterwards, and that installer is not installed anymore with `datadog_remote_updates`

**Note: if you previously had `datadog_remote_updates` in your playbook, the old installer package (deb/rpm) will need to be removed manually with `apt remove datadog-installer` or equivalent, before re-running the playbook and waiting for an hour. This is to ensure the old installer is not reporting anymore.**

**For new installations (hosts that never had the old installer standalone deb/rpm package), simply setting `datadog_remote_updates` will make the host eligible for upgrade.**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -499,7 +499,6 @@ workflows:
               ansible_version: ["2_10", "3_4", "4_10"]
               os: ["rocky8", "amazonlinux2023"]
               apm_enabled: ["host", ""]
-              remote_updates: ["true", "false"]
 
       - test_installer_over_pinned
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -259,8 +259,6 @@ jobs:
         default: "ci.ini"
       apm_enabled:
         type: string
-      remote_updates:
-        type: string
     docker:
       - image: datadog/docker-library:ansible_<<parameters.os>>_<<parameters.ansible_version>>
     steps:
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -220,8 +220,19 @@ jobs:
       - run: cd /tmp/system-tests/lib-injection/build/docker/python/dd-lib-python-init-test-django && sudo docker build -t system-tests/local .
       - run: sudo docker run -d --name test-app-python -p 5985:18080 system-tests/local:latest
       - run: curl --retry 10 --retry-max-time 30 --retry-all-errors localhost:5985
-      # verify that the emitted trace is in trace-agent log
```
tbavelier, line 223: This log no longer exists, so we use a more reliable way to check injection was successful by checking if the trace-agent indeed received spans

ci_test/install_agent_7_apm_python.yaml
```diff
@@ -7,4 +7,4 @@
     datadog_api_key: "11111111111111111111111111111111"
     datadog_agent_major_version: 7
     datadog_apm_instrumentation_enabled: "all"
-    datadog_apm_instrumentation_languages: ["python"]
```
tbavelier, line 10: This was a typo, `datadog_apm_instrumentation_languages` does not exist, the parameter is with `_libraries`

BaptisteFoy (approved):

swang392 (approved):
🚀

PR #673, opened 2025-10-02
remove pycache directory during upgrades

arbll:
This PR removes `/opt/datadog-agent/python-scripts/__pycache__` similarly to other installation methods.

### DIFF
--- a/tasks/pkg-debian.yml
+++ b/tasks/pkg-debian.yml
@@ -116,6 +116,12 @@
     update_cache: true
   when: (datadog_apt_repo | length > 0) and (not ansible_check_mode)
 
+- name: Remove /opt/datadog-agent/python-scripts/__pycache__
+  ansible.builtin.file:
+    path: /opt/datadog-agent/python-scripts/__pycache__
+    state: absent
+  when: not agent_datadog_skip_install
+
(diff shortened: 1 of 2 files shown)


tasks/pkg-debian.yml
```diff
@@ -116,6 +116,12 @@
     update_cache: true
   when: (datadog_apt_repo | length > 0) and (not ansible_check_mode)
 
+- name: Remove /opt/datadog-agent/python-scripts/__pycache__
+  ansible.builtin.file:
+    path: /opt/datadog-agent/python-scripts/__pycache__
+    state: absent
+  when: not agent_datadog_skip_install
```
tbavelier:
```suggestion
  when: not agent_datadog_skip_install
  ignore_errors: true
```
Let's add a safeguard to not block the whole playbook if for some reason this task fails

tbavelier (changes requested):
This only changes the debian installation. I believe this should also be applied to suse/redhat installs e.g. in `tasks/main.yml` before the individual debian/suse/rhel install tasks (since they are the ones doing the upgrade) to avoid duplicating it, since this affects .65 to .70

tbavelier (approved):

PR #682, opened 2025-11-10
[AGENTONB-2675] Remove deprecated deb/rpm installer and support SSI using `install-ssi.sh` script

tbavelier:
## What does this PR do ?
* Removes installation and configuration of the deprecated standalone `datadog-installer` deb/rpm package: mostly 7beaf9b4963980a58a47d59abefdd31ba03fd445
    * Removes associated tests from `.circleci/config.yml`
* Adds a Linux task to remove deprecated installs if preview/older version: a436bf7cee6cbcf5fa1a28bfc30028cab4a2e609 and b591578f0f8c68b95ced6ead1450913208cdd448
* Same as install script, use `install-ssi.sh` script for APM SSI

## Files help
* Removing older deprecated installs to support smooth upgrade: `tasks/remove-legacy-installer.yml`
* SSI support with the script: `tasks/apm-single-step-instrumentation.yml` and `tasks/main.yml`
* All other files: removes installation and configuration of datadog-installer

## QA
* Done manually: ensured Agent is installed/restarted if config changes, ensured APM SSI is enabled by reviewing `datadog-installer` status
* Existing tests for APM injection show no regression

## Notes:
On some distros like AL2023, with Python3, we hit the "classic":
```
fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "The Python 2 yum module is needed for this module. If you require Python 3 support use the `dnf` Ansible module instead."}
```
when using `ansible.builtin.package`. While we could use the specific correct module (example in 090196c7be6f91d0854dd62f2370b4c4bcac50dc), it adds complexity/is ugly and more error-prone. Instead, let's ignore errors and treat it as best-effort -> it will fail on AL2023 depending on the Ansible version, but it's fine as we treat it as best-effort removal, it should impact less and less customers (only those who were in preview)


.circleci/config.yml
tbavelier, line 1: Auto YAML formatting and removal of standalone installer tests

defaults/main.yml
```diff
@@ -181,33 +179,27 @@ datadog_windows_config_root: "{{ ansible_facts.env['ProgramData'] }}\\Datadog"
...
 
+# Advanced users - Configure datadog-installer to support advanced scenarios
+datadog_installer_registry: ""
+datadog_installer_auth: ""
+datadog_installer_version: ""
+datadog_apm_inject_version: ""
+
 # Configure APM host injection. Possible values are: "all", "host" or "docker".
-datadog_apm_instrumentation_enabled: ""
+# datadog_apm_instrumentation_enabled: ""
```
tbavelier, line 189: APM SSI is opt-in: users should specifically set a value. There is validation later down the line

defaults/main.yml
```diff
@@ -181,33 +179,27 @@ datadog_windows_config_root: "{{ ansible_facts.env['ProgramData'] }}\\Datadog"
...
 
 # List of APM libraries with pinned versions to install if host or Docker injection is enabled.
 # Example: `["java:1.23.0", "python:2.0.9"]`
 # You can see the available values in our official docs:
 # https://docs.datadoghq.com/tracing/trace_collection/library_injection_local
 datadog_apm_instrumentation_libraries: []
 
-# Used to send telemetry data to datadog on installation
-datadog_apm_telemetry_endpoint: "https://instrumentation-telemetry-intake.{{ datadog_site | default('datadoghq.com') }}/api/v2/apmtelemetry"
```
tbavelier, line 194: No tasks to send installer telemetry, so let's remove this un-used variable

defaults/main.yml
```diff
@@ -181,33 +179,27 @@ datadog_windows_config_root: "{{ ansible_facts.env['ProgramData'] }}\\Datadog"
...
 # datadog_infrastructure_mode: ""
 
-# Registry, auth, and version for the datadog installer
-datadog_installer_registry: ""
-datadog_installer_auth: ""
-datadog_installer_version: ""
-installer_registry_config: {}
-
-# Version for the datadog apm inject package
-datadog_apm_inject_version: ""
```
tbavelier, line 209: Moved higher up

tasks/_agent-linux-macos-shared.yml
```diff
@@ -169,3 +169,4 @@
     owner: "{{ agent_dd_user }}"
     group: "{{ agent_dd_group }}"
     mode: "0644"
+  when: datadog_apm_instrumentation_enabled is defined
```
tbavelier, line 172: Required by removing the default empty string value for `datadog_apm_instrumentation_enabled` while it was being used in telemetry rendering. `install.json` is only used when APM SSI is enabled, so no need to create the file if not enabled.

tasks/pkg-redhat.yml
```diff
@@ -91,16 +91,9 @@
 
     - name: Set plain includepkgs variable
       ansible.builtin.set_fact:
-        agent_datadog_includepkgs: "{{ datadog_agent_flavor }}
-         {{ agent_dd_apm_install_pkgs | default([], true) | join(' ') }}"
```
tbavelier, line 95: This should have been removed as part of a previous clean-up PR when removing legacy legacy APM libraries (pre-OCI, we'd install them as deb/rpm)

cswatt (approved):

README.md
```diff
@@ -46,27 +46,6 @@ To deploy the Datadog Agent on hosts, add the Datadog role and your API key to y
 
 The API key is required and its absence causes the role to fail. If you want to provide it through another way, outside of Ansible's control, specify a placeholder key and substitute the key at a later point.
 
-### Air-gapped environments
-
-To install Datadog in an air-gapped environment using a specific registry and mirrored images, use the Datadog Ansible collection along with the `datadog_installer_registry`, `datadog_installer_auth`, and `agent_datadog_config` variables. While Datadog recommends fetching `latest`, `datadog_installer_version` and `datadog_apm_inject_version` can be provided to pin versions.
```
fanny-jiang, line 51: If we're removing this documentation, does that mean the new OCI installer does not support air-gapped environments?

fanny-jiang (approved):

README.md
tbavelier: it should, but the setup on customer side is slightly different, they also need to host the ssi script now. Moreover, considering it's advanced/complex, I elected to remove it from the readme

PR #683, opened 2025-11-17
[AGENTONB-2495][Windows] Provide users a way to skip and force reinstall of Datadog Agent

tbavelier:
### What does this PR do

Introduces for Windows two new parameters publicly available:
* `datadog_windows_skip_install`: skip install-related tasks 
* `datadog_windows_force_reinstall`: force reinstallation:
    * Currently, we do not allow downgrades, only same version change to enable NPM. This PR allows to now downgrade by having the role uninstall the installed version before installing the new one, going over the MSI limitation when providing the parameter

### Motivation

Currently, we have internal facts:
* `agent_datadog_skip_install`: skip install if we detect the desired version is already installed. On Windows, unlike Linux, not specifying a version always re-installs latest since we can't know unlike on Linux when using package managers
* `agent_datadog_force_reinstall`: force reinstall if NPM is enabled https://github.com/DataDog/ansible-datadog/blob/8218ea5696054ed8d5f521e9804d00a0fe74fa5c/tasks/pkg-windows-opts.yml#L117-L120

Now, this logic can be overridden explicitly by users providing these values in their playbook.

### QA

Already tested on Windows:
* `datadog_windows_skip_install`: on a bare VM (no Agent installed), use a playbook with `datadog_windows_skip_install: true` and ensure the Agent doesn't get installed, it will error out later since `C:\ProgramData\Datadog` does not exist
    * Then, install the Agent by removing the variable, and re-add it, ensure the latest does not get re-downloaded
    
    | Without `datadog_windows_skip_install` | With `datadog_windows_skip_install` |
    |----------|------|
    | image | image |
    
* `datadog_windows_force_reinstall`: provide `datadog_windows_force_reinstall: true` and `datadog_agent_version` and ensure the same version gets re-installed / provide a lower `datadog_agent_version` and it should downgrade instead of erroring. Example below starting from `7.72.2` and trying to downgrade to `7.70.1`

| Without `datadog_windows_force_reinstall` | With `datadog_windows_force_reinstall` |
|----------|------|
| image | image |

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ These variables provide additional configuration during the installation of the 
 | `datadog_windows_ddagentuser_password`      | The password used to create the user and/or register the service (Windows only).|
 | `datadog_apply_windows_614_fix`             | Whether or not to download and apply file referenced by `datadog_windows_614_fix_script_url` (Windows only). See https://dtdg.co/win-614-fix for more details. You can set this to `false` assuming your hosts aren't running Datadog Agent 6.14.\*.|
 | `datadog_windows_skip_install`              | Set to `true` to force the role to skip Windows Agent installation even if the role detects that installation is required (Windows only).|
-| `datadog_windows_force_reinstall`           | Set to `true` to force the role to reinstall the Windows Agent regardless of the automatic feature detection logic (Windows only).|
+| `datadog_windows_force_reinstall`           | Set to `true` to force the role to reinstall the Windows Agent. This option takes precedence over `datadog_windows_skip_install` (Windows only). |
 | `datadog_macos_user`                        | The name of the user to run Agent under. The user has to exist, it won't be created automatically. Defaults to `ansible_user` (macOS only).|
 | `datadog_macos_download_url`                | Override the URL to download the DMG installer from (macOS only).|
 | `datadog_apm_instrumentation_enabled`       | Configure APM instrumentation. Possible values are: <br/> - `host`: Both the Agent and your services are running on a host. <br/> - `docker`: The Agent and your services are running in separate Docker containers on the same host.<br/>- `all`: Supports all the previous scenarios for `host` and `docker` at the same time.|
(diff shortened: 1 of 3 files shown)


urseberry (approved):

README.md
```diff
@@ -89,6 +89,8 @@ These variables provide additional configuration during the installation of the
 | `datadog_windows_ddagentuser_name`          | The name of Windows user to create/use, in the format `<domain>\<user>` (Windows only).|
 | `datadog_windows_ddagentuser_password`      | The password used to create the user and/or register the service (Windows only).|
 | `datadog_apply_windows_614_fix`             | Whether or not to download and apply file referenced by `datadog_windows_614_fix_script_url` (Windows only). See https://dtdg.co/win-614-fix for more details. You can set this to `false` assuming your hosts aren't running Datadog Agent 6.14.\*.|
+| `datadog_windows_skip_install`              | Set to `true` to force the role to skip Windows Agent installation even if the role detects that installation is required (Windows only).|
+| `datadog_windows_force_reinstall`           | Set to `true` to force the role to reinstall the Windows Agent regardless of the automatic feature detection logic (Windows only).|
```
swang392: Could we add somewhere that `datadog_windows_force_reinstall` takes precedence over `datadog_windows_skip_install`? Since if both are set to true, `agent_datadog_skip_install` gets set to false in that step

tasks/pkg-windows.yml
```diff
@@ -76,11 +118,16 @@
...
   ansible.windows.win_package:
-    path: "{{ agent_download_msi_result.dest }}"
+    product_id: "{{ item }}"
     state: absent
-  when: not agent_datadog_skip_install and agent_datadog_force_reinstall
+  loop: "{{ agent_datadog_windows_installed_product_ids }}"
+  when:
+    - not agent_datadog_skip_install
+    - agent_datadog_force_reinstall
+    - agent_datadog_windows_installed_product_ids is defined
```
swang392, line 129: I'm not sure if this is a case we need to take into account, but what happens if we try to force reinstall on a clean host without the Agent present (ie `agent_datadog_windows_installed_product_ids` isn't defined)? Should we fail gracefully or install the Agent anyway?

swang392 (approved):

tasks/pkg-windows.yml
tbavelier: We should install the Agent anyway (current behaviour from this PR), force_reinstall means force reinstalling regardless of the Agent being present or not imho.
```console
TASK [../../ansible-datadog : Look up installed Datadog Agent product ids when forcing reinstall] ******************************************************************************************************
ok: [win]

TASK [../../ansible-datadog : Cache installed product ids when available] ******************************************************************************************************************************
skipping: [win]
[...]
TASK [../../ansible-datadog : Uninstall installed Datadog Agent to update optional features or downgrade or because requested] *************************************************************************
skipping: [win]
```
basically, we skip removal because there's no product IDs

PR #690, opened 2025-12-26
Write system probe config whenever config is defined

### LINKED ISSUE #668: [BUG] Since role  version 4.28.0 network_config configuration ignored
Duplicating [datadog.dd collection Issue](https://github.com/ansible-collections/Datadog/issues/132) here as I think this is more relevant to the role than the collection.

In the DD Collection version 5.8.0 the role version was bumped to 4.28.0.

> **Environment details (Operating System, Cloud provider, etc):**
> Linux
> 
> **Describe what happened:**
> Any changes to the `network_config` block are ignored. I've tried to disable an enabled `network_config`, but it stays enabled. The issue was found on version 6.2.0. After that, I downgraded the collection to version 5.6.0, and it works. I've tried to upgrade to v5.7.0 and 5.8.0, and it has been reproduced on 5.8.0.
> 
> **Describe what you expected:**
> network_config should be able to enable or disable via Ansible.
> 
> **Steps to reproduce the issue:**
> Try to change `network_config` statement `enabled: false` / `enabled: true`

khewonc:
Write the system-probe.yaml config whenever a config is defined (even if it is `enabled: false`)

Fixes https://github.com/DataDog/ansible-datadog/issues/668

Tested by setting `network_config.enabled: true` then setting `network_config.enabled: false` in the latest released ansible version and checking system-probe.yaml wasn't updated, then updating to this branch and checking that system-probe.yaml was updated

### DIFF
--- a/tasks/agent-linux.yml
+++ b/tasks/agent-linux.yml
@@ -93,7 +93,19 @@
     mode: "0640"
     owner: root
     group: "{{ datadog_group }}"
-  when: datadog_manage_config and not datadog_skip_running_check and agent_datadog_sysprobe_enabled
+  when: datadog_manage_config and not datadog_skip_running_check and (system_probe_config is defined or network_config is defined or
+    service_monitoring_config is defined or runtime_security_config is defined or system_probe_other_config is defined)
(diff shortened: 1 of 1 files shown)


fanny-jiang (approved):

ghost:
That was quick! Thank you.

PR #707, opened 2026-03-19
[FA] - Delete default remote_updates false

Hitsuji-M:
Having the `remote_updates: false` is not the intended pattern. We want users to install new agents with the capability to update their agents remotely through Fleet Automation.

Removing a default will allow new agents to automatically enable it.


Stanislas167 (approved):

swang392 (approved):

.circleci/config.yml
```diff
@@ -0,0 +1,492 @@
+---
```
tedkahwaji: 👋  can we remove this file; we are getting rid of circleci in favor of github actions

Hitsuji-M: Sure ! I was sure it got deleted when I rebased my PR sorry !

